PR #11, opened 2022-09-30
Rendering system

Snow106:
Added Scene object to help manage and render objects
Rendering system developed with ECS implementation
Created shader generation for rendering


KevinAlexV (dismissed):
Looking good on the naming conventions! (unless I'm missing something?)

Only request is if you can add comments describing the function's uses. A brief description would be appreciated, especially in each class.

Engine/src/Engine/Components.h
```diff
@@ -1,11 +1,73 @@
...
+	ColorComponent() = default;
+	ColorComponent(glm::vec4 color)
+		: color(color) {}
+
+	glm::vec4 color;
+};
+
+
+	//Not component, just container for vertex attribute data format
+	struct VertexAttribute
```
ReleaseTheSpice, line 49: Why are these functions indented?

Engine/src/Engine/Scene.cpp
```diff
@@ -9,4 +21,177 @@ namespace Engine
         entity.addComponent<TagComponent>(tag);
         return entity;
     }
+
+	void Scene::onRuntimeStart()
```
ReleaseTheSpice, line 25: Why is this indented?

Engine/src/Engine/Scene.cpp
```diff
@@ -9,4 +21,177 @@ namespace Engine
...
+		}
+
+		onRuntimeStop();
+	}
+
+	void Scene::onRuntimeStop()
+	{
+		glfwTerminate();
+
+		const auto view = getEntities<const VerticesComponent>();
```
ReleaseTheSpice, line 47: const variables should be uppercase according to style guidelines

KevinAlexV (changes requested):
I feel like cpp files should contain the comments about a function's uses, like the h files. Perhaps a more detailed overview of the function in the cpp, while the header just provides a general overview and return type?

I suggest detailed comments follow the beat of 'This does W by doing X and Y, thus returning Z.', while general comments are 'This does W, and returns Z'.


These are the files I think need comments:

ShaderGenerator.h & .cpp functions
Scene.cpp functions
Entity.h & cpp
components.h

Scene.h is a great example of header file comments, from what I was describing above!

Engine/src/Engine/ShaderGenerator.cpp
```diff
@@ -0,0 +1,64 @@
+#include "ShaderGenerator.h"
+#include "Log.h"
+#include "glad/glad.h"
+
+namespace Engine
+{
+	ShaderGenerator::ShaderGenerator(const char* vertexSource, const char* fragmentSource)
```
ReleaseTheSpice, line 9: Is this the generic class for making shaders?  Maybe a couple comments on its purpose and use could help

Engine/src/Engine/ShaderGenerator.cpp
```diff
@@ -0,0 +1,64 @@
...
+		glCompileShader(shaderObj);
+
+		//check for errors
+		if (requestStatus(shaderObj, GL_COMPILE_STATUS) == -1)
+			return -1;
+
+		return shaderObj;
+	}
+
+	int ShaderGenerator::requestStatus(GLuint shaderObj, GLenum status)
```
ReleaseTheSpice: Add comment saying what status is/what you would need it for?

ReleaseTheSpice (commented):
Honestly most of the comments were fine there was just some weird indentation at parts

Engine/src/Engine/Scene.cpp
Snow106: I think the guidelines put it in the context of specifically global consts. The guidelines themselves have examples with const in lower case. Might be worth it to clarify it in guidelines

Engine/src/Engine/ShaderGenerator.cpp
Snow106: I think this function is self-explanatory given that it's being called from genShader.
Since it's private I don't think API explanation is required

KevinAlexV (approved):
I'm personally on the boat of 'Comment everything', even if it's a redundant comment. I view it as if I were explaining this to someone with only a basic knowledge of programming. 

That being said, the comments here are enough for me to glean what's going on. 

I definitely think we should have a discussion about what comments should look like, and how much detail they need to contain (I can provide examples of what I normally do), as I don't like just imposing my own take on comments. 

Looks good either way!

ReleaseTheSpice:
All good

ReleaseTheSpice (approved):

PR #13, opened 2022-10-04
Delta Time implementation

Snow106:
Added to relevant methods in Scene.cpp

### DIFF
--- a/Engine/src/Engine/DeltaTime.h
+++ b/Engine/src/Engine/DeltaTime.h
@@ -1,0 +1,31 @@
+#pragma once
+
+namespace Engine
+{
+	//Container for the DeltaTime used in game.
+	class DeltaTime
+	{
+	public:
+		DeltaTime(double dt = 0)
+		: m_Time(dt)
+		{
+		}
+
+		operator double() const { return m_Time; }
+
+		double getMilliseconds() const { return m_Time * 1000; }
+		double getSeconds() const { return m_Time; }
+
+		//Updates the value of deltaTime.
+		void updateDeltaTime()
+		{
+			const double time = glfwGetTime();
+			m_Time = time - m_previousTime;
+			m_previousTime = time;
+		}
+
+	private:
+		double m_Time;
+		double m_previousTime;
+	};
+}
--- a/Engine/src/Engine/DeltaTime.h
+++ b/Engine/src/Engine/DeltaTime.h
@@ -13,10 +13,13 @@ namespace Engine
 
 		operator double() const { return m_Time; }
 
+		//returns the deltaTime in milliseconds.
 		double getMilliseconds() const { return m_Time * 1000; }
+
+		//returns the deltaTime in seconds.
 		double getSeconds() const { return m_Time; }
 
-		//Updates the value of deltaTime.
+		//Updates the value of deltaTime. Has to be called every frame.
 		void updateDeltaTime()
 		{
 			const double time = glfwGetTime();
--- a/Engine/src/Engine/Scene.cpp
+++ b/Engine/src/Engine/Scene.cpp
@@ -12,6 +12,7 @@
 #include "ShaderGenerator.h"
 #include "SourceGatherer.h"
 #include "Components.h"
+#include "DeltaTime.h"
 
 namespace Engine
 {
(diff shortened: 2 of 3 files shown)


mikesinclair66 (dismissed):
Scene::initializeGL 's block needs to be indented properly, but that's not worth declining this change.

Engine/src/Engine/DeltaTime.h
```diff
@@ -0,0 +1,31 @@
...
+{
+	//Container for the DeltaTime used in game.
+	class DeltaTime
+	{
+	public:
+		DeltaTime(double dt = 0)
+		: m_Time(dt)
+		{
+		}
+
```
KevinAlexV, line 13: What does this do? Do I just get the current time or something?

Engine/src/Engine/DeltaTime.h
```diff
@@ -0,0 +1,31 @@
...
+		operator double() const { return m_Time; }
+
```
KevinAlexV, line 15: If I understand correctly, these return deltatime in milliseconds or seconds, yes?

Engine/src/Engine/Scene.cpp
```diff
@@ -53,9 +55,9 @@ namespace Engine
 		glDeleteProgram(m_programId);
 	}
 
-	void Scene::onRuntimeUpdate()
```
KevinAlexV, line 56: This is just the update function that calls whatever needs updating, correct? So it renders, swap OpenGL buffers, and then polls all events (not sure what events would be included, but that seems more like a comment for that particular function)

KevinAlexV (changes requested):
If what I explained is correct, can it be commented in a similar (and probably concise) way?

Engine/src/Engine/Scene.cpp
Snow106: it's an update function for runtime. It does all those things, but commenting it to that level of detail just creates a lot of maintenance because this function might be modified by other entities and scripting later on.

Snow106: In this case I think descriptive method names are more valuable than a bunch of comments

Engine/src/Engine/DeltaTime.h
Snow106: It's an overloaded operator for implicit conversions.

KevinAlexV: can you provide context on how it's used? I'm not too familiar with this

Engine/src/Engine/Scene.cpp
KevinAlexV: I think in this case, yes, that's true, but perhaps we should at least list what the function should be updating in a comment. Since to me, I feel like this should only update rendering, but if that's the case, I have no idea why poll events is relevant (mostly because I don't know what that function does). 

Thoughts?

KevinAlexV (changes requested):
Added some replies

Engine/src/Engine/DeltaTime.h
Snow106: I'm happy to explain how overloading types in cpp work for anyone that's interested, but I don't think comments in code should explain language features or programming concepts.

Engine/src/Engine/Scene.cpp
Snow106: both glswapbuffers and glfwPollEvents are commented by the library where they're from. It's hard to comment what the function would be updating because this function would be updating anything that needs to be updated. Right now it's only rendering the scene and the subsequent glfw methods, but later on, it might be entity behavior, entity creation/deletion (if we don't have pooling), and so on. How this function works isn't any different than what's already described in the .h

Engine/src/Engine/DeltaTime.h
KevinAlexV: That's fair, I'm just curious how this would be used if it's just returning a double again. But that's probably due to my lack of knowledge. I'll dismiss this for now if you really think no one would glean anything from a comment.

Engine/src/Engine/Scene.cpp
KevinAlexV: Ah, I didn't think about the header comment, my bad.
If so, that should be fine. I'd advise this get a comment if the function becomes more complex, as having people update a few lines when the function is changed should not cause as much overhead as explaining it to someone who doesn't understand it.

KevinAlexV (approved):

ReleaseTheSpice (approved):

PR #15, opened 2022-10-06
Library tests

KevinAlexV:
This is the base resource manager for loading assets and storing them in maps.

Future assets would be handled by the ImportResources class.

Updated Scene.cpp to use this system for shader imports. Can be modified as needed.

Please ask if you have any questions!


Engine/src/Engine/Application.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 #include "Core.h"
+#include "ImportResources.h"
```
Snow106: Probably best to add this to Core.h instead

Snow106: Engine.h*

Engine/src/Engine/ImportResources.h
```diff
@@ -0,0 +1,72 @@
...
+			//Function to return a shader from the hashmap based on a provided name. 
+			std::string returnShader(std::string Name);
+#pragma endregion
+			
+		private:
+
+#pragma region Maps and Vectors
+			
+			//Every file path found under the specified resources folder, 'm_sourcePath'.
+			std::map<std::string, std::string> m_filePaths{};
```
Snow106: Might want to consider std::unordered_map. std::map doesn't hash, and since there is no benefit in keeping things ordered then std::unordered_map should be optimal.

Engine/src/Engine/ImportResources.h
```diff
@@ -0,0 +1,72 @@
...
+		
+		public:
+			
+#pragma region Constructors and Destructors	
+			
+			//Constructor for ImportResources Class. Finds all files under the 'data' path and stores the paths in a map. 
+			ImportResources();
+			
+			//Deconstructor for ImportResources.
+			~ImportResources(){}
```
Snow106: Follow Guideline "Consider the rule of zero". Destructor is not necessary here

Engine/src/Engine/ImportResources.h
```diff
@@ -0,0 +1,72 @@
...
+#include <filesystem>
+#include <map>
+
+
+//Written by Kevin Vilanova, KevinAlexV.
+//Class is intended to import and manage assets and resources used by the game engine.
+namespace Engine
+{
+
+	class ENGINE_API ImportResources
```
Snow106: Just a thought but ImportResources sounds more like a method name rather than a class

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+		{
+			std::string path = m_sourcePaths[i];
+
+			//Using a recursive iterator, check each path under the root paths (m_sourcePaths) and add the file path to m_filePaths.
+			for (std::filesystem::recursive_directory_iterator i(path), end; i != end; ++i)
+			{
+				//File path is the full path of the found folder or file. Filename is the folder/file name, and file extension should contain the file type (to differentiate between file and folder).
+				std::filesystem::path filePath = i->path();
+				std::string fileName = filePath.filename().string();
+				std::string fileExtension = "";
```
Snow106: default initialization of std::string is an empty string. No need to initialize it here

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+				auto it = m_filePaths.find(fileName);
+
+				//If the filename contains a '.' (file extension), save the extension.
+				if (fileName.find(".") != std::string::npos)
+				{
+					fileExtension = fileName.substr(fileName.find_last_of(".") + 1);
+				}
+			
+				//If the file name is not already in the map, and an extension exists, add it to the map.
+				if ((it == m_filePaths.end()) && (fileExtension != ""))
```
Snow106: might be better to call string::empty()

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+				{	
+					m_filePaths.insert(std::pair<std::string, std::string>(fileName, filePath.string()));
+					GE_INFO("[ImportResources] File added to m_filePaths [" + filePath.string() + "]");
```
Snow106: Use GE_CORE_INFO
INFO should be used for game calls (applies for other calls in the file)

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
+#include "ImportResources.h"
+
+namespace Engine
+{
+	
+	//Finds all files under the 'data' path and stores the paths in a map. 
+	ImportResources::ImportResources()
```
Snow106: Since we only have one object that imports resources, I would suggest having a singleton that holds the data for imported resources. In that case, instead of importing the resources in the constructor, you could have a separate method that gathers the data from a relevant place. Might be easier to scale if later on we have different folders with assets (for game or for engine for example)

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+		}
+		else
+		{
+			std::string jsonPath = m_filePaths[Name];
+			std::string extension;
+
+			//GE_INFO("[ImportResources] Path is " + jsonPath);
+
+			//If json path is found (should be loaded into m_filePaths, currently upon initialization), load, store, then return the json.
+			if (jsonPath != "")
```
Snow106: can use std::string.empty()

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+				//File path is the full path of the found folder or file. Filename is the folder/file name, and file extension should contain the file type (to differentiate between file and folder).
+				std::filesystem::path filePath = i->path();
+				std::string fileName = filePath.filename().string();
+				std::string fileExtension = "";
+				
+				//Check if file exists in map.
+				auto it = m_filePaths.find(fileName);
+
+				//If the filename contains a '.' (file extension), save the extension.
+				if (fileName.find(".") != std::string::npos)
```
Snow106: Might want to use single quotes instead of double, since we're looking for a character

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+		{
+			std::string jsonPath = m_filePaths[Name];
+			std::string extension;
+
+			//GE_INFO("[ImportResources] Path is " + jsonPath);
+
+			//If json path is found (should be loaded into m_filePaths, currently upon initialization), load, store, then return the json.
+			if (jsonPath != "")
+			{
+				extension = jsonPath.substr(jsonPath.find_last_of(".") + 1);
```
Snow106: Might want to use single quotes instead of double, since we're looking for a character

Engine/src/Engine/SourceGatherer.h
```diff
@@ -11,13 +11,13 @@ namespace Engine
 	class SourceGatherer
```
Snow106: Class can be deleted

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+		
+			//GE_INFO("[ImportResources] Search Result is " + searchResult);
+
+			//If shader path is found (should be loaded into m_filePaths, currently upon initialization), load, store, then return the shader.
+			if (searchResult != "")
+			{
+				//Ensure the file is a shader file, based on the extension. 
+				extension = searchResult.substr(searchResult.find_last_of(".") + 1);
+
+				if (extension == "fs" || extension == "vs")
```
Snow106: Magic numbers. Might want to make private members than catalogize important extensions

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+		}
+		else 
+		{
+			std::string searchResult = m_filePaths[Name];
+			std::string extension;
+		
+			//GE_INFO("[ImportResources] Search Result is " + searchResult);
+
+			//If shader path is found (should be loaded into m_filePaths, currently upon initialization), load, store, then return the shader.
+			if (searchResult != "")
```
Snow106: string.empty()

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+			{
+				//Ensure the file is a shader file, based on the extension. 
+				extension = searchResult.substr(searchResult.find_last_of(".") + 1);
```
Snow106: single quotes

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+				}
+			}
+		}
+
+		GE_INFO("[ImportResources] " + Name + " not found.");
+		return nullptr;
+	}
+
+	//Based on the provided filename, return the shader source from a map, or load it from the system.
+	std::string ImportResources::returnShader(std::string Name)
```
Snow106: Just my opinion, but I would rename this to getShaderData or something along those lines. Calling a function returnx might be confusing since return is usually a keyword used to exit out of the method.

Engine/src/Engine/Scene.h
```diff
@@ -52,6 +53,8 @@ namespace Engine
         int m_windowWidth{480};
         int m_windowHeight{480};
 
+        ImportResources m_importResources;
```
Snow106: Can be forward declared if you follow the above suggestion of making a singleton

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+	//Finds all files under the 'data' path and stores the paths in a map. 
+	ImportResources::ImportResources() 
+	{
+		//Store all files listed under m_sourcePaths in the m_filePaths map
+		for (int i = 0; i < m_sourcePaths.size(); i++)
+		{
+			std::string path = m_sourcePaths[i];
+
+			//Using a recursive iterator, check each path under the root paths (m_sourcePaths) and add the file path to m_filePaths.
+			for (std::filesystem::recursive_directory_iterator i(path), end; i != end; ++i)
```
Snow106: i is being re-declared here.

Engine/src/Engine/ImportResources.cpp
Snow106: Also just a suggestion, you can instead call the function as:
GE_INFO("[ImportResources] File added to m_filePaths [{0}]", filePath.string());

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+		//If loaded, return the json. Else, load and store the json, then return it.
+		if (it != m_jsons.end())
+		{
+			//Get value based on map key "Name"
+			data = m_jsons.find(Name)->second;
+			
+			GE_INFO("[ImportResources] Previously loaded " + Name + " found.");
+			return data;
+		}
+		else
```
Snow106: Else is redundant here since the if condition returns.

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+					GE_INFO("[ImportResources] File added to m_filePaths [" + filePath.string() + "]");
+				}
+			}
+		}
+	}
+
+#pragma region Load and Return Functions
+	
+	//Based on the provided filename, return the json file from a map, or load it from the system.
+	nlohmann::json ImportResources::returnJson(std::string Name)
```
Snow106: Not part of the review but as far as I can understand, this gets an entire json's data. 
Are there any plans to get specific things from a json? My understanding was that serialization was part of this task, though looking at the trello now maybe I was just confused.

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+		}
+
+		GE_INFO("[ImportResources] " + Name + " not found.");
+		return "";
+	}
+	
+	//Read the source of a file and return it as a string.
+	std::string ImportResources::readSource(std::string sourcePath)
+	{
+		std::string src = "";
```
Snow106: No need for initialization here

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+				}
+			}
+		}
+
+		GE_INFO("[ImportResources] " + Name + " not found.");
+		return "";
+	}
+	
+	//Read the source of a file and return it as a string.
+	std::string ImportResources::readSource(std::string sourcePath)
```
Snow106: Consider making the parameter a const ref, since the method doesn't intend to alter it

Snow106: Same for the above methods

Engine/src/Engine/ImportResources.cpp
```diff
@@ -0,0 +1,155 @@
...
+		nlohmann::json data = nullptr;
+		std::map<std::string, nlohmann::json>::iterator it = m_jsons.find(Name);
+		
+		//If loaded, return the json. Else, load and store the json, then return it.
+		if (it != m_jsons.end())
+		{
+			//Get value based on map key "Name"
+			data = m_jsons.find(Name)->second;
+			
+			GE_INFO("[ImportResources] Previously loaded " + Name + " found.");
```
Snow106: Consider making this a WARN or ERROR

Engine/src/Engine/ImportResources.h
```diff
@@ -0,0 +1,72 @@
...
+			ImportResources();
+			
+			//Deconstructor for ImportResources.
+			~ImportResources(){}
+#pragma endregion
+
+
+#pragma region Return Functions
+			
+			//Function to return a json from the hashmap based on a provided name.
```
Snow106: Might want to specify that it returns an empty string if no Json or Shader is found

Engine/src/Engine/ImportResources.cpp
KevinAlexV: And now I know that's a thing. Updated

Engine/src/Engine/ImportResources.cpp
KevinAlexV: I did not know that's how GE_INFO worked. So the {0} acts as a reference for whatever argument follows, effectively representing its location?

Snow106: Yes. you can pass any number of arguments, where the number inside the braces represents the index of the argument after the initial string. So for example
GE_INFO("{0} {1}, {0} {2}, {0} {1} and {2}!", "hello", "world", "life") would print: "hello world, hello life, hello world and life"

Engine/src/Engine/ImportResources.cpp
KevinAlexV: I'm not sure what the serialization plan would be, but this does just store an entire json's data. What we access from that json depends on future needs, so for now I'm just storing it in a map. Down the line, I can imagine this class being used to export the same json data from this map.

Engine/src/Engine/ImportResources.h
KevinAlexV: It really does, thank you for making me finalize my decision to change it haha

Engine/src/Engine/ImportResources.cpp
KevinAlexV: Would this be worth it? I explored the route before, and each file needs it's own separate processing. I'm not sure how many shader file types we'll be using, but if there are more than 2, then perhaps a variable to store shader extensions might be good.

Engine/src/Engine/ImportResources.cpp
Snow106: sure, I'll make a ticket on trello specifically to serialization then

Engine/src/Engine/Scene.h
```diff
@@ -3,6 +3,7 @@
 #include "entt.h"
 #include "Components.h"
 #include "DeltaTime.h"
+#include "ResourceManager.h"
```
Snow106: include in .cpp

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -0,0 +1,164 @@
...
+			
+		//If shader path is found (should be loaded into m_filePaths, currently upon initialization), load, store, then return the shader.
+		if (shaderPath != m_filePaths.end())
+		{
+			
+			std::string searchResult = m_filePaths[Name];
+			//Ensure the file is a shader file, based on the extension. 
+			std::string extension = searchResult.substr(searchResult.find_last_of('.') + 1);
+
+			if (extension == "fs" || extension == "vs")
```
Snow106: "fs", "vs", and "Json" are still magic variables. Should have these as private members.

Engine/src/Engine/ImportResources.cpp
KevinAlexV: That is really good to know. If you get a chance, can you document that in Log.h?

Engine/src/Engine/ResourceManager.cpp
KevinAlexV: Ah, sorry I misunderstood what a magic variable was. This should be resolved

Engine/src/Engine/ResourceManager.h
```diff
@@ -0,0 +1,107 @@
...
+			
+			#pragma region File Storage Variables
+			
+			//Extensions for all files handled through resource manager.
+			std::string m_jsonFileExt = "json";
+
+			std::vector<std::string> m_shaderFileExts = { "vs", "fs" };
+			#pragma endregion
+
+			#pragma region File Storage Variables
```
Snow106, line 78: I'm going to assume these are indented correctly but are misrepresented in github ?

Snow106 (approved):

Engine/src/Engine/ResourceManager.h
KevinAlexV: Oh, is there a standard for pragma regions? I've never used them, so I'm not sure what the standard is.

Snow106: No I just mean the indenting in the entire file. Some of it seems really pushed forward, but I saw that in other reviews too so I'm assuming it's a GitHub thing

KevinAlexV: Gotcha. Not sure, but they look fine on my end

ReleaseTheSpice (approved):

PR #17, opened 2022-10-07
Audio asset library testing

pike983:
Adding the audio asset library.


out/build/x64-Debug/thirdparty/glfw3/src/glfw3.pc
```diff
@@ -0,0 +1,13 @@
+prefix=C:/University/BTech TERM 3/COMP 8551/EngineProjectRedo/GameEngine/out/install/x64-Debug
```
Snow106: What's the purpose of these files? where these missing from the previous GLFW3 install?

Engine/src/Engine/Audio.h
```diff
@@ -0,0 +1,86 @@
+#pragma once
+#ifndef _AUDIO_H_
+#define _AUDIO_H_
+
+#include <iostream>
+#include "..\irrKlang\irrKlang.h"
+
+// A singleton for controlling sounds throughout the application.
+// Contacts: James
+//
```
Snow106: missing namespace

Engine/src/Engine/Audio.h
```diff
@@ -0,0 +1,86 @@
+#pragma once
+#ifndef _AUDIO_H_
+#define _AUDIO_H_
+
+#include <iostream>
+#include "..\irrKlang\irrKlang.h"
```
Snow106: shouldn't use relative paths to include irrKlang. This should already be available from CMake as an additional include

Engine/src/Engine/Audio.h
```diff
@@ -0,0 +1,86 @@
...
+
+};
+
+
+#pragma region ExampleCode
+
+// For further examples of how to use the ISoundEngine and
+// ISound, view: https://www.ambiera.com/irrklang/docu/index.html
+
+/*
```
Snow106: Coding standards: don't use /* comments.

Engine/src/Engine/Audio.cpp
```diff
@@ -0,0 +1,69 @@
...
+// More information on sound effects: https://www.ambiera.com/irrklang/docu/index.html#effects
+// 
+ISound* AudioPlayerSingleton::playSound(const char* soundPath, bool loop, bool startPaused, bool useSoundEffects)
+{
+	ISound* theSound = engine->play2D(soundPath, loop, startPaused, true, ESM_AUTO_DETECT, useSoundEffects);
+	if (theSound != 0)
+	{
+		return theSound;
+	}
+	return NULL;
```
Snow106: use nullptr

Engine/src/Engine/Audio.cpp
```diff
@@ -0,0 +1,69 @@
+#include "Audio.h"
+
+using namespace std;
+using namespace irrklang;
+
+AudioPlayerSingleton* AudioPlayerSingleton::instance;
+
+AudioPlayerSingleton* AudioPlayerSingleton::getInstance()
+{
+	if (instance == 0)
```
Snow106: Should check for nullptr instead of 0

Engine/src/Engine/Audio.cpp
```diff
@@ -0,0 +1,69 @@
...
+// Plays a sound, takes a path to the sound as a const char, a bool if the sound should loop,
+// a bool if the sound should be initialized paused or play on initialization, and a bool for if
+// sound effects should be used. Sound effects need to be added after using this function using
+// the returned ISound*.
+// More information on sound effects: https://www.ambiera.com/irrklang/docu/index.html#effects
+// 
+ISound* AudioPlayerSingleton::playSound(const char* soundPath, bool loop, bool startPaused, bool useSoundEffects)
+{
+	ISound* theSound = engine->play2D(soundPath, loop, startPaused, true, ESM_AUTO_DETECT, useSoundEffects);
+	if (theSound != 0)
```
Snow106, line 28: does the library actually return a 0 or a nullptr if there's an error? If this is to keep parity with the library then it's probably fine

Engine/src/Engine/Audio.cpp
```diff
@@ -0,0 +1,69 @@
...
+	{
+		return theSound;
+	}
+	return NULL;
+	//TODO: implement alternate media folder search if multiple media folders
```
Snow106, line 33: It's probably better to leave the TODO here and create a task in Trello with the relevant code to be added

Engine/src/Engine/Audio.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// Plays a sound in 3D space, takes a const char* path to the sound, 
+// an irrklang::vec3df position (irrklang::vec3df positionName(x,y,z)),
+// a bool if the sound should loop, a bool if the sound should initialize paused,
+// and a bool if the sound should use sound effects. Sound effects need to be added after
+// using this funtion using the returned ISound*.
+// More information on sound effects: https://www.ambiera.com/irrklang/docu/index.html#effects
+// 
+ISound* AudioPlayerSingleton::play3DSound(const char* soundPath, vec3df sound3DPosition, bool loop, bool beginPaused, bool useSoundEffects)
```
Snow106, line 44: is pass by value ok here? (for the vec3def)

Engine/src/Engine/Audio.h
```diff
@@ -0,0 +1,86 @@
...
+#include <iostream>
+#include "..\irrKlang\irrKlang.h"
+
+// A singleton for controlling sounds throughout the application.
+// Contacts: James
+//
+class AudioPlayerSingleton
+{
+private:
+	static AudioPlayerSingleton* instance;
```
Snow106: guidelines: m_instance

Snow106 (changes requested):
Everything in out/build/x64-Debug should not be merged into develop.
add relevant filepaths to .gitignore, though I would note that it seems a lot of these were added because certain configurations were changed from VS. Configs in this project should be done through Cmake, reach out if you need help with that

Engine/src/Engine/Audio.cpp
pike983: Yeah, I think it should be. The returned ISound lets the position be changed later.

Engine/src/Engine/Audio.cpp
pike983: The library returns a 0 if it can't play the sound.

Engine/src/Engine/Audio.h
```diff
@@ -4,80 +4,83 @@
...
 #include "irrKlang.h"
 
-// A singleton for controlling sounds throughout the application.
-// Contacts: James
-//
-class AudioPlayerSingleton
-{
-private:
-	static AudioPlayerSingleton* instance;
+namespace IAudio {
```
Snow106: Bracket in new line

Snow106: also namespace should still be Engine

Snow106 (changes requested):

Snow106 (approved):

mikesinclair66 (dismissed):
src/Engine/Audio.cpp can't use "using namespace example_name";
guidelines

Snow106:
> src/Engine/Audio.cpp can't use "using namespace example_name"; guidelines

While I personally perfer leaving "using" out of .cpp files, the guidelines specifically allow the use of using in implementation files "Writing using namespace in an implementation file is fine though."

mikesinclair66 (approved):

PR #18, opened 2022-10-07
[Feature] Scripting

Snow106:
- Added Script Component. This Component needs to be added to an Entity, and then it needs to Bind with a ScriptableBehavior derived class.
- Added SciptableBehavior. Works similarly to Monobahvior in unity, though a lot less robust.
- Added missing dll exports/imports in Scene.

I was planning on creating a custom script but since we're missing Input I couldn't think of a script I could add that would actually be useful in the long run of the engine. Ideally I was going to make some camera controls for Camera or something like that.


KevinAlexV (dismissed):
Looks good to me - comments are great! 

Might be good to have a set of instructions on how to create scripts somewhere (we have your screenshot, but it will likely get buried at some point).

Snow106:
> !

I mentioned it in the description of the PR but my idea was to make a custom script to do something... but because we don't have input setup yet I couldn't think of anything that would stay in the engine long term. Open to ideas

Engine/src/Engine/Components.h
```diff
@@ -72,4 +72,26 @@ namespace Engine
...
+
+	//Defines a component to create custom script actions using ScriptableBehavior
+	struct ScriptComponent
+	{
+		ScriptableBehavior* m_instance{ nullptr };
+
+		std::function<ScriptableBehavior*()> instantiateScript;
+		std::function<void()> destroyScript;
+
+		//Links the Script Component to a scpefic ScriptableBehavior
```
roughleythomas: spelling: specific

Engine/src/Engine/Scripts/ScriptableBehavior.h
```diff
@@ -0,0 +1,56 @@
...
+	protected:
+		//gets called once after Entities are serialized and created, in Scene::createEntities().
+		virtual void onCreate() {}
+
+		//gets called before the entity is destroyed.
+		//currently unused since we are not destroying entities until the game is exited.
+		virtual void onDestroy() {}
+
+		//Runs on a loop before rendering
+		virtual void onUpdate(DeltaTime dt) {}
```
roughleythomas: const DeltaTime& ?

Engine/src/Engine/Scripts/ScriptableBehavior.h
```diff
@@ -0,0 +1,56 @@
...
+
+		//Runs after all other Entity onUpdates have been called and after rendering
+		virtual void onLateUpdate(DeltaTime dt) {}
```
roughleythomas: const DeltaTime& ?

roughleythomas (changes requested):

roughleythomas (approved):

KevinAlexV (approved):
In regards to your comment, I completely forgot about an input system, and if a team isn't already working on it, we should get some folks to take a look at that. Some way to read input, and a way to update components with the most recent input (or something that allows us to send input to functions that need to read the most recent input). 

Guess that's for our next meeting, if we're all booked.

PR #19, opened 2022-10-09
Thread pool

ReleaseTheSpice:


KevinAlexV (commented):
Few things:
1. Can I get a brief explanation about how you would queue up a job in the JobQueue? Let's say I want to accomplish a task such as loading textures (I have 10 I want to store), and I think it can be done on a thread. Would I be able to use this in order to accomplish said task? If so, how? 
2. What should we consider moving forward with multithreading? Does each new class need it's own mutex locks to be thread-safe, or does this system handle it? Maybe I'm just confused about how this affects the rest of the engine.
3. Are the files under [out/build/x64-Debug] essential? They seem like files that need to be added in the gitignore.

The comments look good, and I have a general understanding of what's going on - my major confusion is just about how this affects everything else.

Snow106 (changes requested):
/Out/ folder should not be submitted in develop since it's visual studio generated. Please remove it from the branch

Engine/src/Engine/JobQueue.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include <list>
+#include <mutex>
+#include "ThreadJob.h"
+
+namespace engine_concurrent
```
Snow106: we can just use the engine namespace here (and other files). I don't think there's a need to separate our code into different namespaces at this point

Engine/src/Engine/Thread.h
```diff
@@ -0,0 +1,34 @@
...
+        explicit Thread(int id); // explicit constructor for threads
+
+        // Accessor for job
+        // Returns: the job being performed by the thread
+        // Currently not used by anything, just for debugging
+        ThreadJob* getJob();
+
+        // Method run by the underlying thread, runs whatever is in m_threadJob
+        void run();
+    protected:
```
Snow106: Unused protected field?

Engine/src/Engine/Thread.cpp
```diff
@@ -0,0 +1,51 @@
+#include "Thread.h"
+#include "JobQueue.h"
+#include "Log.h"
+#include <thread>
```
Snow106: already included in thread.h

Engine/src/Engine/Thread.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <iostream>
+
+using namespace engine_concurrent;
+
+// THREADS
+#pragma region Threads
+
+Thread::Thread(int id) : 
+	m_threadJob(nullptr),
+	m_thread (new std::thread(&Thread::run, this)),
```
Snow106, line 14: Is this a memory leak? I don't see a destructor for thread; when does this new thread get deleted?

Engine/src/Engine/Thread.cpp
```diff
@@ -0,0 +1,51 @@
...
+// Returns the job being performed by the thread
+ThreadJob* Thread::getJob()
+{
+	return m_threadJob;
+}
+
+// Method to run m_threadJob
+void Thread::run()
+{
+	std::chrono::milliseconds sleepTime(500);
```
Snow106: Magic variable here. Consider making it a private member

Engine/src/Engine/Thread.cpp
```diff
@@ -0,0 +1,51 @@
...
+//			GE_CORE_TRACE("Thread {d} working on {s}", m_id, m_threadJob->getName());
+			m_threadJob->run(); // Run the thread job
+			//GE_CORE_TRACE("Thread {0} finished {1}", m_id, m_threadJob->getName());
+			delete m_threadJob;
+//			GE_CORE_TRACE("Thread {0} grabbing another job", m_id);
+			m_threadJob = queue->getJob();
+//			GE_CORE_TRACE("Thread {0} grabbed {1}", m_id, m_threadJob == nullptr ? "nothing" : m_threadJob->getName());
+			//m_threadJob = nullptr; // Then go back to no job
+		}
+	} while (true); // This 'true' can be changed if we ever need to add a condition for threads to run
```
Snow106: would a function overload make sense here? If we passed in a condition for a parameter, and just default that condition to true?

Engine/src/Engine/ThreadJob.h
```diff
@@ -0,0 +1,48 @@
...
+		// Thread jobs should not be copied, able to be cloned or assigned.  These statements prevent this:
+		ThreadJob(ThreadJob& other) = delete;
+		void operator=(ThreadJob& other) = delete;
+		
+		// Called by the thread to run the passed in function with the passed in parameters
+		void run();
+		// Accessors to return job properties
+		const Priority& getPriority();
+		const std::string& getName();
+	protected:
```
Snow106: empty protected field

Engine/src/Engine/ThreadJob.h
```diff
@@ -0,0 +1,48 @@
...
+		//
+
+	private:
+		// EntryPoint run by thread, contains the function provided by the caller
+		EntryPoint* m_pEntryPoint;
+		void* m_param;          // Parameters of entrypoint
+        Priority m_priority;    // Priority of job
```
Snow106, line 43: weird indent here? I'm going to assume it's github again but best to double check

Engine/src/Engine/ThreadJob.cpp
```diff
@@ -0,0 +1,26 @@
...
+	: m_pEntryPoint(entryPoint),    // Entry point is a pointer to the functions to be run
+	m_priority(priority), 
+	m_param(param),                 // Parameters to be passed to the entry point function
+	m_jobName(name)
+{
+}
+
+void ThreadJob::run()
+{
+	m_pEntryPoint(m_param); // Run the passed-in functions with the passed params
```
Snow106, line 15: A double pointer function here is a bit confusing here. I would suggest considering std::function here for clarity

Engine/src/Engine/ThreadPool.h
```diff
@@ -0,0 +1,34 @@
...
+	{
+	public:
+		// Prevent copying or assigning this class multiple times to keep it a singleton
+		ThreadPool(ThreadPool& other) = delete; 
+		void operator=(const ThreadPool&) = delete;
+
+		// Returning a pointer reference prevents duplicating the pool to make it singleton
+		static ThreadPool* getInstance(); 
+
+	protected:
```
Snow106: empty protected?

Engine/src/Engine/ThreadPool.h
```diff
@@ -0,0 +1,34 @@
...
+
+	private:
+		// Private c'tor to prevent accidental construction
+		ThreadPool(); // Redefine constructor to ensure threadpool can't be instantiated outside the class
+
+		// Mutex ensuring that only one instance of the class is created
+		static std::mutex mutex; // 'Static' mutex so the class can use this mutex without it being bound to an instance
```
Snow106: style guidelines: m_mutex

Engine/src/Engine/ThreadPool.h
```diff
@@ -0,0 +1,34 @@
...
+		// Singleton instance
+		static ThreadPool* tpInstance;
```
Snow106: style guidelines: m_tpInstance

Engine/src/Engine/ThreadPool.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+// THE JOB OF THREAD POOL IS JUST TO MAKE ALL THE THREADS and maybe more things later 
+
+// Constructor for thread pool, is run once at the start and shouldn't be run again
+//make 1 thread for each core, -1 to leave space for the main thread
+ThreadPool::ThreadPool() : m_threads(std::thread::hardware_concurrency()-1)
+{
+    for (int i = 0; i < m_threads.size(); i++)
+    {
+        m_threads[i] = new Thread(i);
```
Snow106, line 17: I don't see Thread getting deleted. Possible memory leak?

Snow106: Also is there a reason to call "new" for the wrapper class here? Can this be created on the stack?

Engine/src/Engine/JobQueue.h
```diff
@@ -0,0 +1,41 @@
...
+
+		// Method to get the singleton instance
+		// Returns: the singleton instance
+		static JobQueue* getInstance();
+	protected:
+	private:
+		// Private c'tor to prevent accidental construction
+		JobQueue();
+
+		static JobQueue* instance;          // The instance
```
Snow106: missing prefix for these variables

Engine/src/Engine/JobQueue.h
```diff
@@ -0,0 +1,41 @@
...
+		void postJob(ThreadJob* job);
+
+		// Grab a job from the queue, will be called by threads when they need a job
+		// Returns: the job at the front of the queue
+		ThreadJob* getJob();
+
+		// Method to get the singleton instance
+		// Returns: the singleton instance
+		static JobQueue* getInstance();
+	protected:
```
Snow106: empty protected?

Engine/src/Engine/JobQueue.cpp
```diff
@@ -0,0 +1,54 @@
...
+	std::unique_lock lock(queueMutex);
+	//std::condition_variable
+	if (m_jobList->empty())
+	{
+		m_jobList->push_back(job);
+		return;
+	}
+	// Iterate over the job list
+	auto it = m_jobList->begin();
+	for (; (it != m_jobList->end() // If we're not at the end of JobList,
```
Snow106, line 34: How does this behave if the jobList is already full? would it not overwrite the last job in the list?

Engine/src/Engine/JobQueue.cpp
```diff
@@ -0,0 +1,54 @@
...
+	{
+		instance = new JobQueue();
+	}
+	return instance;
+}
+
+JobQueue::JobQueue() : m_jobList(new std::list<ThreadJob*>()) {}
+JobQueue::~JobQueue() { delete m_jobList; }
+
+void JobQueue::postJob(ThreadJob* job)
```
Snow106, line 23: just to be sure here I would suggest changing the return type here to bool, where the return is a success state of the job being posted

Engine/src/Engine/JobQueue.cpp
```diff
@@ -0,0 +1,54 @@
...
+	std::unique_lock lock(instanceMutex);
+	if (instance == nullptr)
+	{
+		instance = new JobQueue();
+	}
+	return instance;
+}
+
+JobQueue::JobQueue() : m_jobList(new std::list<ThreadJob*>()) {}
+JobQueue::~JobQueue() { delete m_jobList; }
```
Snow106, line 21: since destructor is being declared here I would suggest deleting the move/copy/assign like the other classes in the PR

Snow106 (changes requested):

Engine/src/Engine/Thread.h
ReleaseTheSpice: Yes, just left in for organization
I want to familiarize myself with c++ header/class structure so I left in all the templates for things so I know where to put it if I need a protected variable in the future
I'd prefer to keep it but can remove it if that's best practice

KevinAlexV (dismissed):
Read me looks good, thank you! Just make sure to resolve the conflict with gitignore (seems to have something to do with out/build)

THREAD_README.txt
```diff
@@ -0,0 +1,62 @@
+THREAD README
```
Snow106: This feels like it should be added to the repository wiki

KevinAlexV: Probably but we can make that a separate PR

Snow106 (dismissed):

KevinAlexV (dismissed):
Great work guys!

Engine/src/Engine/ThreadPool.cpp
roughleythomas: Fixed: added d'tor

KevinAlexV (approved):

Snow106 (approved):

PR #21, opened 2022-10-14
Sprite rendering

KevinAlexV:
Finalized Sprite Rendering. 

Todo: 
- Adjust screen scaling so sprites scale with window size, and the canvas is not just increased. 

- Create new class to contain GLuint IDs for VBO, IBO, and VAOs of commonly used objects (in this case, sprites). This would allow us to only load them once, and not each time we create a sprite (don't have to redefine verts and such)

- Add sprite sheet handling via ResourceManager. Look into BitBlt, as suggested by Borna

Let me know if something needs more explaining, or if any adjustments are necessary!


Snow106:
Please add the TODOs to trello if you haven't already.

Snow106:
I would suggest renaming Bluedemise2.png to something more descriptive, like bluedemise_splash or bluedemise_text or whatever we use the text version for

Game/Data/Shaders/Fill.vs
```diff
@@ -1,10 +1,17 @@
 #version 330 core
 
 layout (location = 0) in vec3 pos;
+layout(location = 1) in vec2 vertexUV;
+//layout (location = 2) in vec2 aTexCoord;
```
Snow106: can we get rid of commented code if it's not explaining anything?

Engine/src/Engine/Entity.h
```diff
@@ -18,26 +18,31 @@ namespace Engine
 		Entity(const Entity& other) = default;
 		inline entt::entity getHandle() { return m_entityHandle; }
 		const std::string& getName() { return getComponent<TagComponent>().tag; }
-
-		//Functions to help with managing entities' components
+		
+#pragma region Entity Component Management	
+		
+		//A template for adding components to an entity.
```
Snow106, line 24: Not requestion a change just providing my opinion:
All of these functions are named well enough that they're descriptive without the need for a comment. These comments feel redundant

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -1,8 +1,12 @@
-#include "ResourceManager.h"
+//Needed to define stbi class, for stb related use in this file. Essentially defines code to be used in this file.
+#define STB_IMAGE_IMPLEMENTATION
```
Snow106, line 3: we probably want a guard here

Engine/src/Engine/ResourceManager.h
```diff
@@ -35,73 +47,88 @@ namespace Engine
...
 		protected:
-			#pragma region Constructors
+			#pragma region Constructors & Deconstructors
 			
-						//Constructor for ResourceManager Class. Finds all files under the 'data' path and stores the paths in a map. 
-						ResourceManager();
+			//Constructor for ResourceManager Class. Finds all files under the 'data' path and stores the paths in a map. 
+			ResourceManager();
+
+			//Deconstructor
```
Snow106: typo: destructor

Engine/src/Engine/ResourceManager.h
```diff
@@ -35,73 +47,88 @@ namespace Engine
...
 			
 			//static lock m_mutex is the lock to be used for static functions. Otherwise, m_functionlock should be utilized.
 			static std::mutex m_mutex;
 			std::mutex m_functionLock;
 			
-			#pragma region File Storage Variables
+			#pragma region File Extension Variables
 			
 			//Extensions for all files handled through resource manager.
 			std::string m_jsonFileExt = "json";
```
Snow106: I know this wasn't put in this PR but guidelines do say to initialize objects with {} instead of = when possible, I personally forgot about that guideline before. Might want to refactor other initializations in this file.

Engine/src/Engine/ResourceManager.h
```diff
@@ -3,6 +3,8 @@
 #include "Core.h"
 #include "json.h"
 #include "Log.h"
+#include "glad/glad.h"
+#include "GLFW/glfw3.h"
```
Snow106: glad and glfw3 types can be forward declared

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -71,6 +86,20 @@ namespace Engine
 #pragma endregion
 	
 #pragma region Get Functions
+	//Obtain icon at filepath stored in this class, then return icon as GLFW image.
+	void ResourceManager::setIcon(GLFWwindow& window)
```
Snow106: Method name feels misleading. You might expect to set the icon for a given entity or object, but it only sets the icon for the application window. (I know it specifically requests a window parameter, but it could still be misinterpreted).
Maybe rename to setApplicationIcon

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -115,6 +144,84 @@ namespace Engine
 		return nullptr;
 	}
 
+	//Based on the provided filename, return the GLuint ID for the texture stored within a map, or load it from the system.
+	//Currently only supports 2D Textures, but can be changed using texType.
+	GLuint ResourceManager::getTexture(const std::string& Name)
+	{
+		//std::lock_guard<std::mutex> lock(m_functionLock);
```
Snow106: should this be commented out?

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -115,6 +144,84 @@ namespace Engine
 		return nullptr;
 	}
 
+	//Based on the provided filename, return the GLuint ID for the texture stored within a map, or load it from the system.
+	//Currently only supports 2D Textures, but can be changed using texType.
+	GLuint ResourceManager::getTexture(const std::string& Name)
```
Snow106: guidelines: param variable should be lower case

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -71,6 +86,20 @@ namespace Engine
...
+
+		//Create a GLFW image and load the icon into it.
+		images[0].pixels = stbi_load(m_iconPath.c_str(), &images[0].width, &images[0].height, 0, 4);
+		
+		glfwSetWindowIcon(&window, 1, images);
+		stbi_image_free(images[0].pixels);
+	}
+
 	//Based on the provided filename, return the json file from a map, or load it from the system.
 	nlohmann::json ResourceManager::getJsonData(const std::string& Name)
```
Snow106: guidelines: param variable should be lower case

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -115,6 +144,84 @@ namespace Engine
...
 
+	//Based on the provided filename, return the GLuint ID for the texture stored within a map, or load it from the system.
+	//Currently only supports 2D Textures, but can be changed using texType.
+	GLuint ResourceManager::getTexture(const std::string& Name)
+	{
+		//std::lock_guard<std::mutex> lock(m_functionLock);
+
+		//Initilaze the return value, and create a map iterator to check if the data is already loaded.
+		GLuint data;
+		std::unordered_map<std::string, GLuint>::iterator it = m_textures.find(Name);
```
Snow106: I'm not sure if I understand why an iterator is needed in a hashmap. Could we instead check if `m_texture.contains(Name)` and continue from there?

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -115,6 +144,84 @@ namespace Engine
...
+			//Configures the way the texture repeasts
+			glTexParameteri(texType, GL_TEXTURE_WRAP_S, GL_MIRRORED_REPEAT);
+			glTexParameteri(texType, GL_TEXTURE_WRAP_T, GL_MIRRORED_REPEAT);
+			
+			// Extra lines in case you choose to use GL_CLAMP_TO_BORDER
+			// float flatColor[] = {1.0f, 1.0f, 1.0f, 1.0f};
+			// glTexParameterfv(GL_TEXTURE_2D, GL_TEXTURE_BORDER_COLOR, flatColor);
+			
+			// Assigns the image to the OpenGL Texture object based on the number of RGB components (if 4, then alpha value is present).
+			if (img.numComponents == 3)
```
Snow106: could we use the enum here instead of having 3/4? (magic variables)

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -185,5 +292,47 @@ namespace Engine
 		
 		return src;
 	}
+
+	//Based on the provided filename, load the image from the system and return the imageData struct contain all data provided by stbi_load. 
+	//NOTE: After using the image, you MUST use stbi_image_free(); in order to free the memory of the image.
+	ResourceManager::ImageData ResourceManager::readImageData(const std::string& Name)
```
Snow106: style guidelines: parameter should be lowecased

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -185,5 +292,47 @@ namespace Engine
...
+	{
+		//Initilaze the return value, and create a map iterator to check if the data is already loaded.
+		ResourceManager::ImageData data = { nullptr, 0, 0, 0 };
+
+		//Create an iterator to check if the file exists in the map (done since using m_filePaths[Name] will create a new entry if it doesn't exist).
+		std::unordered_map<std::string, std::string>::iterator imagePath = m_filePaths.find(Name);
```
Snow106: iterator feels a bit weird to find specific data, though I might be misunderstanding these methods

Engine/src/Engine/ResourceManager.h
```diff
@@ -35,73 +47,88 @@ namespace Engine
...
+
+			
+			//Function to return a shader (formatted as a string with newlines to seperate GSLS code) from the hashmap based on a provided name. Returns an empty string if no shader is found.
 			std::string getShaderData(const std::string& Name);
 			#pragma endregion
 
 		
 		protected:
-			#pragma region Constructors
+			#pragma region Constructors & Deconstructors
```
Snow106: typo: destructors

Engine/src/Engine/Scene.h
```diff
@@ -57,7 +74,15 @@ namespace Engine
 
         DeltaTime m_deltaTime{0};
 
+		//GL IDs for various objects. 
         GLuint m_programId;
+        GLuint m_spriteVBO;
+        GLuint m_spriteVAO;
+        GLuint m_spriteIBO;
+
+        bool createdVBO = false;
```
Snow106: guidelines: initialize values with {}

Engine/src/Engine/Scene.h
```diff
@@ -27,28 +37,35 @@ namespace Engine
...
-        template<typename... Components>
-        auto getEntities()
-        {
-            return m_registry.view<Components...>();
-        }
+        #pragma endregion
 
     private:
+
+        #pragma region OpenGL Scene Management
```
Snow106, line 44: Side note but did we confirm #pragma regions work in other IDEs?

Engine/src/Engine/Scene.cpp
```diff
@@ -107,32 +110,71 @@ namespace Engine
 	{
 		glClear(GL_COLOR_BUFFER_BIT);
 
-		auto view = getEntities<const TransformComponent, const VerticesComponent, const ColorComponent>();
+		//Get entities that contain transform & vertices & color components,
```
Snow106: These comments feel to redundant. getEntities is already descriptive enough, we should trust our engineers to understand what these assignments are saying.

Engine/src/Engine/Scene.cpp
```diff
@@ -107,32 +110,71 @@ namespace Engine
...
 
-			glDrawElements(GL_TRIANGLES, 3, GL_UNSIGNED_INT, nullptr);
+			glDrawElements(GL_TRIANGLES, vertices.numIndices, GL_UNSIGNED_INT, nullptr);
 		}
+
+		//For each updatable sprite entity (with transform, vertices, and color components), draw them.
+		for (auto [entity, transform, vertices, texture, color] : sprites.each())
+		{
+			//Get GLuint for texture, and bind texture for rendering
+			glBindTexture(GL_TEXTURE_2D, texture.texID);
```
Snow106, line 145: Really cool how you made this work! You made texture loading really efficient 😎

Engine/src/Engine/Scene.cpp
```diff
@@ -107,32 +110,71 @@ namespace Engine
...
+
+			glm::mat4 mvp = updateMVP(transform, vm, pm);
+
+			GLuint colorUniformID = glGetUniformLocation(m_programId, "col");
```
Snow106: I think since you made updateMVP, it would be valuable to make a similar function here for color. That should avoid code repetition here with the above for loop

Engine/src/Engine/Scene.cpp
```diff
@@ -107,32 +110,71 @@ namespace Engine
...
 		auto cameraView = getEntities<const CameraComponent>();
 		const auto camera = m_registry.get<CameraComponent>(cameraView.back());
 		glm::mat4 pm = glm::ortho(-camera.viewport.x / 2, camera.viewport.x / 2,
 			-camera.viewport.y / 2, camera.viewport.y / 2, camera.nearZ, camera.farZ);
 		glm::mat4 vm = glm::translate(glm::mat4(1.f), glm::vec3(0, 0, -10.f)); //position of camera in world-space
 
-		for (auto [entity, transform, vertices, color] : view.each())
+		//Render all entities
+		//For each updatable solidObj entity (with transform, vertices, and color components), draw them.
+		for (auto [entity, transform, vertices, color] : solidObj.each())
```
Snow106, line 126: just a suggestion, but I think for clarity you could define solidObj and spirtes each above their respective for loop. Tat should make it clearer what each view's intended use is.

Engine/src/Engine/Scene.cpp
```diff
@@ -185,47 +251,104 @@ namespace Engine
...
-		glGenVertexArrays(1, &vc.vaoID);
-		glBindVertexArray(vc.vaoID);
-
-		glGenBuffers(1, &vc.vboID);
-		glBindBuffer(GL_ARRAY_BUFFER, vc.vboID);
-
-    	//Buffer data
-		glBufferData(GL_ARRAY_BUFFER, sizeof(triangleVertices), triangleVertices, GL_STATIC_DRAW);
-		glEnableVertexAttribArray(0);
+		vc.vaoID = getSpriteVAO();
```
Snow106, line 345: this looks much neater now. gj!

Snow106 (changes requested):

Engine/src/Engine/Scene.cpp
KevinAlexV: Honestly, these comments are more for me than anything. My brain registers the green text relatively easily, so I use it as a way to guide myself along chunks of code.
Despite them being redundant, I don't see much of a downside in keeping them. Just for me at the very least

Engine/src/Engine/Scene.h
KevinAlexV: No we haven't. That's perhaps a @ReleaseTheSpice question, as I recall him mentioning using a different IDE. Might be misremembering though

Engine/src/Engine/ResourceManager.cpp
KevinAlexV: The main reason for the iterator is to ensure it's a valid variable. I tried a few different ways to error check (ensure it's not null), and this seemed to be the best solution. I'm open to suggestions

Engine/src/Engine/ResourceManager.cpp
KevinAlexV: Contains doesn't exist on hashmaps

Engine/src/Engine/ResourceManager.cpp
KevinAlexV: This was something for @roughleythomas to look at. I left it there since I'm not sure how we needed to lock it. Mind commenting Thomas?

Engine/src/Engine/ResourceManager.h
KevinAlexV: Sorry, I'm not too sure how to go about that. Maybe I'm not remembering what forward declaration is.

Engine/src/Engine/ResourceManager.cpp
KevinAlexV: A guard..?

Engine/src/Engine/Entity.h
KevinAlexV: See other thread

Engine/src/Engine/ResourceManager.cpp
KevinAlexV: Can we use an Enum for ints? At the very least I'll create an int for this comparison.

Engine/src/Engine/ResourceManager.cpp
Snow106: ah .contains seems to be a c++20 method. My bad here then, all good

Snow106 (dismissed):

Engine/src/Engine/Components.h
```diff
@@ -13,6 +13,7 @@ namespace Engine
 		std::string tag;
 	};
 	
+	//A component for storing the matrices of a camera, and distance/fov.
 	struct CameraComponent 
 	{
 		CameraComponent(float f, glm::mat4 proj, glm::vec2 view, float fZ, float nZ)
```
roughleythomas: Please make input variable names clearer

roughleythomas (changes requested):

Engine/src/Engine/Components.h
KevinAlexV: What specifically should be changed? I didn't write this code, so I'm not too sure what should be altered. @Snow106 @Hevano

KevinAlexV: Ok, I updated this to the best of my knowledge. I'll add a trello ticket to determine a component standard that we can look at later for updating all this.

roughleythomas (approved):

Snow106 (approved):

PR #22, opened 2022-10-19
Feature/jsonexporting

ReleaseTheSpice:


Game/Data/test/test2.json
```diff
@@ -0,0 +1 @@
+{"num":2,"root":{"children":[{"children":[{"name":"child1.1"},{"name":"child1.2"}],"name":"child1"},{"name":"child2","num":3}],"name":"root"}}
```
Snow106: I would suggest removing this test folder and leaving some json files for the unit test team.
Maybe remove these from the PR and create a trello tricket to add json unit tests

Engine/src/Engine/Application.cpp
```diff
@@ -1,10 +1,46 @@
 #include "Application.h"
 #include "Scene.h"
+#include "ResourceManager.h"
```
Snow106: I understand these changes are here for the purposes of testing but they should not make it into develop

Engine/src/Engine/ResourceManager.h
```diff
@@ -37,13 +37,34 @@ namespace Engine
 		//Singletons should not be assignable, this is to prevent that.
 		void operator=(const ResourceManager&) = delete;
 
+
+		//Function to return a shader from the hashmap based on a provided name. Returns an empty string if no shader is found.
+		std::string getShaderData(const std::string& Name);
+
+		//Function to save provided json data to a provided file. Will save to "Data/data.json" if file name/path are not specified.
+		void saveJsonFile(nlohmann::json data, std::string fileName = "data", std::string path = "Data/");
```
Snow106: I'm worried if we should have a default fileName here. It seems like it could lead to very easy accidental overwrites

Engine/src/Engine/ResourceManager.h
```diff
@@ -37,13 +37,34 @@ namespace Engine
...
 		//This retrieves a pointer to the current instance of ResourceManager. If it doesn't exist, then one will be created and returned.
 		static ResourceManager* getInstance();
 
+
 		#pragma endregion
 
 		#pragma region Set Functions
 
+		//The static instance of ResourceManager that will be returned if the resource manager is requested.
+		static ResourceManager* m_pinstance;
```
Snow106: This seems to be duplicate code

Snow106 (changes requested):
The branch isn't building.

Engine/src/Engine/Application.cpp
```diff
@@ -1,10 +1,46 @@
...
 
 namespace Engine
 {
 	void Application::Run()
 	{	
+		nlohmann::json data = {
```
KevinAlexV: This seems more like a test json. If you verified that it works, I imagine this should be removed.

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -89,6 +89,26 @@ namespace Engine
 		}
 	}
 
+	//Save provided json object to a provided file. Defaults file name and path to "Data/data.json"
+	//If the provided file name does not exist, it will be created. Otherwise, the existing file will be overwritten.
+	void ResourceManager::saveJsonFile(nlohmann::json data, std::string fileName, std::string path)
```
KevinAlexV, line 94: I don't mind leaving this till later, but we should probably have some default json location so the 'path' doesn't need to be specified in parameters.

Engine/src/Engine/ResourceManager.h
KevinAlexV: I agree. I do think the default path is smart, but it may be worth changing that too down the line, based on my other comment.

Engine/src/Engine/ResourceManager.h
```diff
@@ -37,13 +37,34 @@ namespace Engine
 		//Singletons should not be assignable, this is to prevent that.
 		void operator=(const ResourceManager&) = delete;
 
+
+		//Function to return a shader from the hashmap based on a provided name. Returns an empty string if no shader is found.
+		std::string getShaderData(const std::string& Name);
+
+		//Function to save provided json data to a provided file. Will save to "Data/data.json" if file name/path are not specified.
+		void saveJsonFile(nlohmann::json data, std::string fileName = "data", std::string path = "Data/");
+		#pragma endregion
```
KevinAlexV: Should remove this. The end region is on the next few lines.

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -89,6 +89,26 @@ namespace Engine
...
+		std::ofstream fileStream(fileWithPath);	// Create a file stream and open the file
+		if (fileStream.is_open())
+		{
+			fileStream << data;					// Replace existing file content with the updated prevData
+			fileStream.close();					// Close the file stream so it doesn't hang open
+		}
+		else
+		{
+			// This should never happen but is left in for debugging purposes.
+			GE_CORE_ERROR("[ResourceManager] The file stream broke!", fileWithPath);
```
KevinAlexV: I think this error is inaccurate. If a filestream isn't open, that just means the file couldn't be read/opened. The file stream would only break on line 102, if at all (not sure I've seen that happen though). Maybe adjust the message?

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -89,6 +89,26 @@ namespace Engine
 		}
 	}
 
+	//Save provided json object to a provided file. Defaults file name and path to "Data/data.json"
```
KevinAlexV: Great explanation! Just adjust the 'provided file' to 'provided filename' to avoid confusion.

KevinAlexV (changes requested):
Few things that should be adjusted, but overall seems to be good!

You have now become the point of contact for jsons haha

Game/Data/test/test2.json
ReleaseTheSpice: Test folder has now been removed

Engine/src/Engine/Application.cpp
ReleaseTheSpice: Includes for testing resource manager have been removed.
This feels a little inconsistent though as I removed similar includes after testing Threads,  yet they were re-added..

Engine/src/Engine/ResourceManager.h
ReleaseTheSpice: If you say so
It has now been removed

Engine/src/Engine/ResourceManager.h
ReleaseTheSpice: Duplicate from where?  Either this has already been resolved or github is pulling things

Engine/src/Engine/Application.cpp
ReleaseTheSpice: whoops

Engine/src/Engine/ResourceManager.cpp
ReleaseTheSpice: Default json location is already set to Data/ in the 'Game' folder

Engine/src/Engine/ResourceManager.cpp
ReleaseTheSpice: updated it to be more accurate

Engine/src/Engine/ResourceManager.cpp
ReleaseTheSpice: good point, updated

KevinAlexV (approved):
No more comments here!

mikesinclair66:
Would have put saveFilePath and saveJsonFile in same function if it were myself, but looks good

Snow106 (approved):

PR #23, opened 2022-10-19
Feature/imgui implementation

Aiden-Lonquist:
skeleton of the GUI with all of the windows created.


Engine/src/Engine/UserInterface.h
```diff
@@ -0,0 +1,59 @@
...
+#include <map>
+#include <string>
+
+#include "imgui.h"
+#include "imgui_impl_opengl3.h"
+#include "imgui_impl_glfw.h"
+
+#include "glm/glm.hpp"
+
+class UserInterface
```
Snow106: mssing namespace

Engine/src/Engine/MainMenu.cpp
```diff
@@ -0,0 +1,73 @@
...
+	{
+		//sets the font colour to be white
+		s_style->Colors[ImGuiCol_Text] = white;
+
+		//Do not use else-if statements here otherwise
+		//the menu titles will not render correctly
+		//creates a drop down menu where if x is true (open), display y
+		if (ImGui::BeginMenu("File"))
+		{
+			if (ImGui::MenuItem("New", nullptr, nullptr))
```
KevinAlexV: These are nullptr because we have no functionality for them yet, correct?

Engine/src/Engine/MainMenu.cpp
```diff
@@ -0,0 +1,73 @@
+#include "MainMenu.h"
+
+void MainMenu::show()
```
KevinAlexV: This is surprisingly clean to me, love it!

Engine/src/Engine/Scene.h
```diff
@@ -3,6 +3,13 @@
 #include "entt.h"
 #include "Components.h"
 
+//Added for UI
+#include <array>
+#include "ExplorerPanel.h"
+#include "EntitiesPanel.h"
+#include "ComponentsPanel.h"
+#include "MainMenu.h"
+
```
KevinAlexV, line 14: I think these can be forward declared so that the headers are only included in the cpp. 
If you can't though, then feel free to disregard this.

Engine/src/Engine/UserInterface.cpp
```diff
@@ -0,0 +1,101 @@
+#include "UserInterface.h"
+
+ImGuiIO* UserInterface::s_IO = nullptr;
+ImGuiStyle* UserInterface::s_style = nullptr;
```
KevinAlexV: s_ ? Is this a naming convention I wasn't aware of?

Engine/src/Engine/UserInterface.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+std::map<std::string, ImFont*> UserInterface::s_fonts;
+
+//define the ImGUI vec4s for each colour
+ImVec4 UserInterface::red = ImVec4(1.0f, 0.0f, 0.0f, 1.0f);
+ImVec4 UserInterface::green = ImVec4(0.0f, 1.0f, 0.0f, 1.0f);
+ImVec4 UserInterface::blue = ImVec4(0.0f, 0.0f, 1.0f, 1.0f);
+ImVec4 UserInterface::white = ImVec4(1.0f, 1.0f, 1.0f, 1.0f);
+
+//initializes a GLFWwindow that can be used with ImGUI
```
KevinAlexV: This comment seems inaccurate, unless I'm misunderstanding. 
Wouldn't this function initialize the UI and attach it to the provided GLFWwindow?

Engine/src/Engine/UserInterface.cpp
```diff
@@ -0,0 +1,101 @@
...
+{
+	ImGui::CreateContext();
+	ImGui_ImplOpenGL3_Init("#version 330");
+	ImGui_ImplGlfw_InitForOpenGL(window, true);
+
+	//TODO - Reference for style
+	//https://www.unknowncheats.me/forum/c-and-c-/189635-imgui-style-settings.html
+	s_style = &ImGui::GetStyle();
+
+	s_IO = &ImGui::GetIO();
```
KevinAlexV: What's IO?

Engine/src/Engine/UserInterface.cpp
```diff
@@ -0,0 +1,101 @@
...
+	ImGui::NewFrame();
+}
+
+void UserInterface::endUI()
+{
+	ImGui::Render();
+	ImGui_ImplOpenGL3_RenderDrawData(ImGui::GetDrawData());
+}
+
+//loads a font, returns true if the font is loaded correctly
```
KevinAlexV: Future consideration: Do these provide UI elements we can use in the game, and not just the engine? If so that may save us some work.

Engine/src/Engine/UserInterface.cpp
```diff
@@ -0,0 +1,101 @@
...
+//defines ImGUI spacing 
+void UserInterface::spacing(int amount)
+{
+	for (int i = 0; i < amount; i++)
+	{
+		ImGui::Spacing();
+	}
+}
+
+//creates the title section above the UI element
```
KevinAlexV: Not sure I understand this. Does that imply every UI element has a 'title section'? Or is this only for menu items?

Engine/src/Engine/UserInterface.h
```diff
@@ -0,0 +1,59 @@
...
+
+	//initialize takes in a GLFWwindow as a parameter and initializes a window to be used with ImGUI
+	static bool initialize(GLFWwindow* window);
+
+	//set position takes in a vec2 and is used to set the position of the window
+	void setPosition(const glm::uvec2& position);
+	void setDimension(const glm::uvec2& dimension);
+	void setButtonDimension(const glm::uvec2& dimension);
+
+	void spacing(int amount);
```
KevinAlexV: "spacing" just adds space around a particular UI element?

KevinAlexV (changes requested):
Few notes if you all can review them. I'll re-review this afterward, but it's very clean, nice work!

Engine/src/Engine/UserInterface.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#include <map>
+#include <string>
+
+#include "imgui.h"
+#include "imgui_impl_opengl3.h"
```
Snow106, line 6: Unused include

Engine/src/Engine/UserInterface.h
```diff
@@ -0,0 +1,59 @@
...
+	static ImVec4 white;
+
+	//initialize start and end functions
+	static void startUI();
+	static void endUI();
+
+	//loadFont takes in a string file name, float font size, and string tag and returns true if the font is valid
+	static bool loadFont(const std::string& fontFilename, float fontSize, const std::string& tag);
+
+	virtual ~UserInterface() = 0 {}
```
Snow106: no need for the braces here, can end with a ; after the 0 ( just like show() )

Engine/src/Engine/UserInterface.h
Snow106: naming here sounds like a member variable. I would suggest renaming to "setSpacing" or something similar

Engine/src/Engine/UserInterface.cpp
Snow106: We don't have any standard for static variables. Prefixing them with an s_ is fine, though since it's being added here we should also add it to our standards for upcoming PRs

Engine/src/Engine/UserInterface.h
```diff
@@ -0,0 +1,59 @@
...
+
+	virtual void show() = 0;
+	
+	void shutdown();
+
+protected:
+
+	static ImGuiIO* s_IO;
+	static ImGuiStyle* s_style;
+	static std::map<std::string, ImFont*> s_fonts;
```
Snow106: Is there an advantage for map over unordered_map here?

Engine/src/Engine/UserInterface.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+//initializes a GLFWwindow that can be used with ImGUI
+bool UserInterface::initialize(GLFWwindow* window)
+{
+	ImGui::CreateContext();
+	ImGui_ImplOpenGL3_Init("#version 330");
+	ImGui_ImplGlfw_InitForOpenGL(window, true);
+
+	//TODO - Reference for style
+	//https://www.unknowncheats.me/forum/c-and-c-/189635-imgui-style-settings.html
```
Snow106: this link should be added to a trello task instead so we can later change the style.
It will be easier to keep track of that way

Engine/src/Engine/Scene.h
Snow106: These are all non pointer members so they can't be forward declared

Engine/src/Engine/Scene.cpp
```diff
@@ -30,6 +30,27 @@ namespace Engine
 		glfwSwapInterval(1);
 		glClearColor(0.1f, 0.1f, 0.1f, 1);
 
+		//initialize the window for UI
+		UserInterface::initialize(m_window);
+
+		const int menuHeight = 18;
+
+		//set position for the explorer UI element
+		m_explorerPanel.setPosition(glm::uvec2(0, menuHeight));
```
Snow106: I would suggest all of these should be moved to their own method, such as "setEditorWindow" or something similar, for example

Engine/src/Engine/Scene.cpp
```diff
@@ -56,6 +77,23 @@ namespace Engine
 	void Scene::onRuntimeUpdate()
 	{
 		renderScene();
+
+		//start the UI
+		UserInterface::startUI();
```
Snow106: I would say that we should have a separate onEditorRuntime to handle all updates related to editor specifically

Snow106: We can then use the editor window to actually show the render of the game, or just use a regular window if the edit is not launched (such as a game executable for example)

Engine/src/Engine/Scene.cpp
Snow106: Similarly to my other comment, we also probably want a specific onEditorCreate() method

Engine/src/Engine/Scene.cpp
```diff
@@ -89,6 +127,11 @@ namespace Engine
 	//clears the window and renders all entities that need to be rendered (those with transform, vertices, color).
 	void Scene::renderScene()
 	{
+		//Added to put triangle view in top/middle of screen
+		glClearColor(0.1f, 0.1f, 0.1f, 1);
```
Snow106: This is where we would then chose to either render in the editor window or in an actual window.

Engine/src/Engine/MainMenu.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+#include "UserInterface.h"
+
+class MainMenu : public UserInterface
+{
+
+public:
+
```
Snow106, line 10: shouldn't we be defining a destructor here? Since the base class has a pure virtual destructor, I would expect to see an override here

Engine/src/Engine/MainMenu.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+#include "UserInterface.h"
+
+class MainMenu : public UserInterface
```
Snow106: missing namespace

Engine/src/Engine/MainMenu.cpp
```diff
@@ -0,0 +1,73 @@
...
+		s_style->Colors[ImGuiCol_Text] = white;
+
+		//Do not use else-if statements here otherwise
+		//the menu titles will not render correctly
+		//creates a drop down menu where if x is true (open), display y
+		if (ImGui::BeginMenu("File"))
+		{
+			if (ImGui::MenuItem("New", nullptr, nullptr))
+			{
+				//Handle menu item...
```
Snow106: I know this might be in process already, but it would be good to have a Trello ticket to keep track of these handles

Engine/src/Engine/ExplorerPanel.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include "UserInterface.h"
+
+class ExplorerPanel : public UserInterface
```
Snow106: missing namespace

Engine/src/Engine/EntitiesPanel.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+#include "UserInterface.h"
+
+class EntitiesPanel : public UserInterface
```
Snow106: missing namespace

Engine/src/Engine/ComponentsPanel.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "UserInterface.h"
+
+class ComponentsPanel : public UserInterface
```
Snow106: missing namespace

Snow106 (changes requested):
Requests aside, the skeleton looks pretty good, good job guys!

Engine/src/Engine/MainMenu.cpp
Blackstaxs: In a manner of speaking, yes, these are nullptr because argument 2 points to a menu shortcut, and argument 3 points to a bool variable to specify if the menu item is selected. Both of these arguments are not being used, hence nullptr.

Engine/src/Engine/UserInterface.cpp
Blackstaxs: I wasn't sure what the standard for static variables was. I assumed s_ since all member variables were m_.

Engine/src/Engine/UserInterface.cpp
Blackstaxs: it does indeed initialize the UI and not the window

Engine/src/Engine/UserInterface.cpp
Blackstaxs: ImGui IO is used later for loading custom fonts.

Engine/src/Engine/UserInterface.cpp
Blackstaxs: Yes, the UI tools are designed to be reusable throughout game and engine.

Engine/src/Engine/UserInterface.cpp
Blackstaxs: This was intended to section off areas of the UI panels with headings.

Engine/src/Engine/UserInterface.h
Blackstaxs: So namespace Engine for all UI classes?

Engine/src/Engine/Scene.cpp
Blackstaxs: so something like explorerPanel.setEditorWindow(position, dimension)?

Engine/src/Engine/Scene.cpp
Blackstaxs: I'm not sure I understand. The UI is part of the core engine, there is no editor window per-se. Where do you want this onEditorRuntime to be?

Engine/src/Engine/UserInterface.cpp
KevinAlexV: I'll add it to our standards then, it makes sense.

Engine/src/Engine/UserInterface.cpp
KevinAlexV: Ok, gotcha. In that case, can you update that comment when you get a chance?

Engine/src/Engine/MainMenu.cpp
KevinAlexV: If you can add a comment explaining the same thing, that would be great!

Engine/src/Engine/UserInterface.cpp
KevinAlexV: can you mention that wherever s_IO is defined via comment?

Engine/src/Engine/UserInterface.cpp
KevinAlexV: Ok, I may follow up with you on that - thanks for mentioning it!

Engine/src/Engine/UserInterface.cpp
KevinAlexV: I see, ok. Might be wise to re-phrase this, but it should be fine. Thanks for clarifying!

Engine/src/Engine/MainMenu.cpp
Snow106: @KevinAlexV The method is properly commented in the library definition.

Engine/src/Engine/UserInterface.h
Snow106: it's part of our guidelines:
"functions and classes should exist in an appropriately named namespace or in a class inside of a namespace"
You can use the "Engine" namespace, just encapsulate the class inside it. This applies for all classes and methods.
For the cpp you might want to include everything inside the same namespace as well, though per our standards you're allowed to use "using".

Engine/src/Engine/Scene.cpp
Snow106: Thinking about it better you probably don't need a method for these, just define everything from ImGui in a separe create method (suggested name onEditorCreate(); That way we can make sure to not call the relevant editor function when an executable of the game is made.

Engine/src/Engine/Scene.cpp
Snow106: Basically we want to separate editor update and game update. That way when we build a game executable, we are not also building the editor as well (just the game window). By separating these onUpdate functions, we can make that process a lot easier later on.

Engine/src/Engine/MainMenu.cpp
Aiden-Lonquist: @KevinAlexV added another comment explaining it more

Engine/src/Engine/UserInterface.cpp
Aiden-Lonquist: @KevinAlexV updated the comment

Engine/src/Engine/UserInterface.cpp
Aiden-Lonquist: @KevinAlexV added comment

Engine/src/Engine/UserInterface.cpp
Aiden-Lonquist: @KevinAlexV changed the phrasing slightly. Hopefully it's more clear.

Engine/src/Engine/UserInterface.h
Aiden-Lonquist: @Snow106 updated this to setSpacing

Engine/src/Engine/UserInterface.h
Aiden-Lonquist: @Snow106 added namespace Engine

Engine/src/Engine/UserInterface.h
Aiden-Lonquist: changing `virtual ~UserInterface() = 0 {}` to `virtual ~UserInterface() = 0;` caused the build to fail. Not sure why.

Engine/src/Engine/MainMenu.h
Aiden-Lonquist: @Snow106 added namespace Engine

Engine/src/Engine/ExplorerPanel.h
Aiden-Lonquist: @Snow106 added namespace Engine

Engine/src/Engine/EntitiesPanel.h
Aiden-Lonquist: @Snow106 added namespace Engine

Engine/src/Engine/ComponentsPanel.h
Aiden-Lonquist: @Snow106 added namespace Engine

Engine/src/Engine/UserInterface.h
Blackstaxs: Removed imgui.h

Engine/src/Engine/UserInterface.h
Blackstaxs: changed it to unordered_map, we don't need it for fonts

Engine/src/Engine/Scene.cpp
Blackstaxs: Created a RenderUI()

Engine/src/Engine/Scene.cpp
Blackstaxs: Created initializeUI()

Engine/src/Engine/UserInterface.h
Snow106: I guess it's because the destructor for UserInterface still needs to be defined since this is being called. Usually " = 0 " at the end of a function indicates that it is a pure virtual function, meaning that the derived classes HAVE to implement the function. Declaring a body in the same base class ignores that functionality. I think in that case we can just use the default constructor (or in our case to follow the rule of 0, just get rid of it altogether).

Snow106 (approved):

KevinAlexV (approved):

PR #25, opened 2022-10-19
Remove thread tests from application

roughleythomas:


Snow106 (dismissed):
LGTM, thanks

Engine/src/Engine/Application.cpp
```diff
@@ -12,65 +12,9 @@ namespace Engine
 {
 	static std::mutex letterProtect;
```
KevinAlexV: Is this necessary to keep in the branch? Not sure what value it has.

KevinAlexV (approved):
Lookin good!

Snow106 (approved):

PR #27, opened 2022-10-25
Audio update

pike983:
Adding alternative media path searching to the audio file.


Engine/src/Engine/Audio.cpp
```diff
@@ -19,38 +21,114 @@ AudioPlayerSingleton* AudioPlayerSingleton::getInstance()
...
+		string filePath = std::filesystem::current_path().parent_path().string() + "\\Engine\\src\\Media\\";
+		filePath += soundNamePath;
+		std::replace(filePath.begin(), filePath.end(), '\\', '/');
+		const char* fullPath = filePath.c_str();
+		ISound* theSound = engine->play2D(fullPath, loop, startPaused, true, ESM_AUTO_DETECT, useSoundEffects);
+		if (theSound != 0)
+		{
+			return theSound;
+		}
+		else
```
Snow106: this else is redundant because of the return

Engine/src/Engine/Audio.cpp
```diff
@@ -19,38 +21,114 @@ AudioPlayerSingleton* AudioPlayerSingleton::getInstance()
...
+		{
+			return nullptr;
+		}
+	}
+	else
```
Snow106: this else is redundant because of the returns

Engine/src/Engine/Audio.cpp
```diff
@@ -19,38 +21,114 @@ AudioPlayerSingleton* AudioPlayerSingleton::getInstance()
...
+			for (const auto& dEntry : std::filesystem::recursive_directory_iterator(filePath2))
+			{
+				if (dEntry.path().filename() == soundNamePath)
+				{
+					ISound* theSound2 = engine->play2D(dEntry.path().string().c_str(), loop, startPaused, true, ESM_AUTO_DETECT, useSoundEffects);
+					if (theSound2 != 0)
+					{
+						return theSound2;
+					}
+					else
```
Snow106: same here

Engine/src/Engine/Audio.cpp
```diff
@@ -19,38 +21,114 @@ AudioPlayerSingleton* AudioPlayerSingleton::getInstance()
...
+			return theSound;
+		}
+		else
+		{
+			string filePath2 = std::filesystem::current_path().parent_path().string() + "\\Game\\src\\Media\\";
+			for (const auto& dEntry : std::filesystem::recursive_directory_iterator(filePath2))
+			{
+				if (dEntry.path().filename() == soundNamePath)
+				{
+					ISound* theSound2 = engine->play2D(dEntry.path().string().c_str(), loop, startPaused, true, ESM_AUTO_DETECT, useSoundEffects);
```
Snow106: This line might be a bit too long, should be shortened according to standards
(Could just show up that way though since it's on github web that looks long. Just double check in IDE)

Engine/src/Engine/Audio.cpp
```diff
@@ -19,38 +21,114 @@ AudioPlayerSingleton* AudioPlayerSingleton::getInstance()
...
+		if (theSound != 0)
+		{
+			return theSound;
+		}
+		else
+		{
+			return nullptr;
+		}
+	}
+	else
```
Snow106: same with the following elses

Engine/src/Engine/Audio.cpp
```diff
@@ -19,38 +21,114 @@ AudioPlayerSingleton* AudioPlayerSingleton::getInstance()
...
+		std::replace(filePath.begin(), filePath.end(), '\\', '/');
+		const char* fullPath = filePath.c_str();
+		ISound* theSound = engine->play2D(fullPath, loop, startPaused, true, ESM_AUTO_DETECT, useSoundEffects);
+		if (theSound != 0)
+		{
+			return theSound;
+		}
+		else
+		{
+			string filePath2 = std::filesystem::current_path().parent_path().string() + "\\Game\\src\\Media\\";
```
Snow106: I suggest having a private member so it's easier to change the dir later on

Engine/src/Engine/Audio.cpp
```diff
@@ -19,38 +21,114 @@ AudioPlayerSingleton* AudioPlayerSingleton::getInstance()
...
+		}
+		else
+		{
+			return nullptr;
+		}
+	}
+	else
 	{
-		return theSound;
+		string filePath = std::filesystem::current_path().parent_path().string() + "\\Engine\\src\\Media\\";
```
Snow106: I suggest having a private member so it's easier to change the dir later on

Engine/src/Engine/Audio.cpp
```diff
@@ -19,38 +21,114 @@ AudioPlayerSingleton* AudioPlayerSingleton::getInstance()
...
+		std::replace(filePath.begin(), filePath.end(), '\\', '/');
+		const char* fullPath = filePath.c_str();
+		ISound* theSound = engine->play3D(fullPath, sound3DPosition, loop, beginPaused, true, irrklang::ESM_AUTO_DETECT, useSoundEffects);
+		if (theSound != 0)
+		{
+			return theSound;
+		}
+		else
+		{
+			string filePath2 = std::filesystem::current_path().parent_path().string() + "\\Game\\src\\Media\\";
```
Snow106: I suggest having a private member so it's easier to change the dir later on

Engine/src/Engine/Audio.h
```diff
@@ -3,6 +3,8 @@
 #define _AUDIO_H_
 
 #include "irrKlang.h"
+#include <string>
```
Snow106, line 6: I don't see either of these includes being used in the header. Consider moving to .cpp

Snow106 (changes requested):
Please add a description to the PR. It will be easier to understand what was integrated later on.

mikesinclair66 (approved):

Snow106 (approved):
Nice work!

PR #32, opened 2022-10-28
[BUG] Post-Processing Assets

Snow106:
We weren't adding assets to the build folder, causing the Game.exe to crash

This creates a copy of the Asset directory in Game and puts it in the folder


KevinAlexV (dismissed):

ReleaseTheSpice (changes requested):
Everything looks good, just make sure to also update the comment for the saveJsonFile in the implementation too, on line 95 of ResourceManager.cpp to be /Assets instead of /Data

KevinAlexV (approved):

ReleaseTheSpice (approved):

PR #33, opened 2022-10-29
Deserialization

Snow106:
* Added skeleton for deserialization.
* Methods to serialize scenes and their entities
* sample scene included.
* Changed Scene creation from Engine to Project
* Added name member to scene. Doesn't do anything, just nice to have for future
* Fixed some misc bugs I encountered.
* Lost my sanity a little bit


Engine/src/Engine/Serializer.cpp
```diff
@@ -0,0 +1,181 @@
...
+			{
+				GE_CORE_FATAL("Unable to deserialize entity {0}", entity["tag"]);
+				GE_CORE_FATAL("A component does not contain the \"name\" attribute");
+				return false;
+			}
+
+			auto name = component["name"].get<std::string>();
+
+			//check each component type individually
+			if (name == "CameraComponent")
```
Hevano: This can be changed to a switch statement for performance. Since c++ doesn't allow switch statements with strings, we could use a map with an enum to replicate the same behaviour.

Engine/src/Engine/Scene.cpp
```diff
@@ -25,12 +25,6 @@ namespace Engine
 #pragma region Runtime Functions
 	void Scene::onRuntimeStart()
 	{
-		if (!initializeGL()) return;
```
Hevano, line 28: Since initializeGL is private, this change essentially means that only friends of Scene can properly initialize scenes. If deserialization is the only way to initialize new scenes, this should be fine. Either way, this decision should likely be noted in a comment in either in Scene constructor or Scene::initializeGL.

Engine/src/Engine/Serializer.cpp
```diff
@@ -0,0 +1,181 @@
...
+		
+			else if (name == "ColorComponent")
+			{
+				out.addComponent<ColorComponent>(component["color"].get<glm::vec4>());
+			}
+		
+			else if (name == "TextureComponent")
+			{
+				std::string texture = component["texName"];
+				auto image = ResourceManager::getInstance()->getTexture(texture);
```
Hevano, line 157: Needs error checking if texture file not found. If no file is found, getTexture will return an empty ImageData object.

Hevano (changes requested):

Engine/src/Engine/Serializer.cpp
Snow106: This is already the case 
image

Engine/src/Engine/Scene.cpp
Snow106: I think this is complex enough now that a Renderer class is warranted but outside of the scope of this PR. I made a trello ticket to track this concern: https://trello.com/c/6cMTYYrU/82-create-renderer-class

Engine/src/Engine/Serializer.cpp
Hevano: I am aware that it logs the error, but in my opinion there should be special handling for a texture not found during the deserialization process. Perhaps we could use default error texture like many other engines do. Otherwise, I believe that failing to find a texture should fail the deserialize function. If loading the texture fails but the empty image data is still passed to the component, the call to bind texture will bind texture 0, which is an unbind call. I feel like this may open the door to a future bug that could be difficult to trace. Having said that, it is not an immediately pressing concern.

Snow106: Yeah I agree with that. It would be great to have more error handling, but I believe that's a rendering issue and possibly out of this PR's scope

Hevano (approved):
Looks great!

Engine/src/Engine/Scene.cpp
KevinAlexV, line 28: So onRuntimeStart will assume GL is always initialized? Just want to make sure this check isn't needed anymore, as I'm not too sure why we would want to remove this

KevinAlexV (approved):
Looking clean!

PR #34, opened 2022-11-01
Sprite sheets

KevinAlexV:
Framework to handle spritesheets. 

Given our decision to utilize SS exclusively for animations, this will only be fully implemented after the animation systems implementation. 

https://trello.com/c/9jsRhfT8/64-resourcemanager-sprite-sheet-handling


Engine/src/Engine/Components.h
```diff
@@ -58,6 +58,21 @@ namespace Engine
 		GLuint texID;
 	};
 
+	//A component containing animation data
+	struct AnimationComponent
```
Snow106, line 62: Make sure to add this into deserialization Serializer.cpp.
LMK if I can help with anything

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -262,6 +222,105 @@ namespace Engine
 	
 	}
 
+	//Based on the provided spritesheet filename, return the vector of GLuints for the sprites stored within a map, or load it from the system.
+	ResourceManager::SpriteSheet ResourceManager::getSpritesheet(const std::string& name, float spriteWidth, float spriteHeight)
+	{
+		//Initilaze the return value, and create a map iterator to check if the data is already loaded.
+		SpriteSheet data {0,0,0,0,0,0,0,0,0,NULL};
```
Snow106: this feels like it should be the default constructor

Engine/src/Engine/ResourceManager.cpp
```diff
@@ -262,6 +222,105 @@ namespace Engine
...
+		ImageData img = readImageData(name);
+
+		//If image data isn't nullptr, store and return data.
+		if (img.image != nullptr)
+		{
+			ImageData spriteSheet = getTexture(name);
+
+			if (spriteWidth == 0 || spriteHeight == 0)
+			{
+				spriteWidth = spriteSheet.width / 3;
```
Snow106: assuming the number of frames in the spritesheet might prove problematic in the future. I would suggest either keeping a variable in the spritesheet struct, or passing it in as a function param.

Snow106: if We're assuming 3 then that can be the default value of either variable/param

Snow106 (changes requested):
Good work! Just a few changes
Side question, why do we need a sprite sheet with no transparency?

Engine/src/Engine/ResourceManager.cpp
KevinAlexV: So with this, typically the frames would be calculated if the function is passed the dimensions of the sprite. 

That if statement is to catch any cases where the dimensions are 0, in which case we would assume our default spritesheet dimensions (the size of the sheet split into 3 columns/rows). 

In terms of defaulting the sprite width/height, that might be hard because that would change based on the size of the spritesheet (and we don't obtain that data until here). 

I'm open to suggestions though

Snow106: I meant defaulting the "3" in the code snippet. Otherwise, it's a magic variable, and it's hard to understand unless we know of the default sprite sheet we would expect

Snow106 (approved):

roughleythomas (approved):

PR #35, opened 2022-11-03
Input system

roughleythomas:
Handles cursor position and mouse input
Hooks for handling key input are available but are not fully tested
Event system to be handled in a separate branch


Engine/src/Engine/InputSystem.cpp
```diff
@@ -0,0 +1,113 @@
+#include "Engine/InputSystem.h"
+
+using namespace Engine;
+
+InputSystem* InputSystem::m_instance = nullptr;
+std::mutex InputSystem::m_instanceMutex;
+
+InputSystem::InputSystem():m_window(nullptr), m_isInit(false)
+{
+	//Moved map initialization to the init function
```
Snow106: Since this was moved might want to just remove this constructor definition

Engine/src/Engine/InputSystem.h
```diff
@@ -0,0 +1,118 @@
+#pragma once
+#include <map>
+#include <mutex>
+#include "GLFW/glfw3.h"
+#include "glm/vec2.hpp"
+
+namespace Engine {
+	/// <summary>
```
Snow106, line 10: Not necessarily against xml style documentation, but it feels inconsistent if this is the only place we have it in the project. Just some thoughts, we could look at standardizing comments later on

Engine/src/Engine/InputSystem.h
```diff
@@ -0,0 +1,118 @@
+#pragma once
+#include <map>
+#include <mutex>
+#include "GLFW/glfw3.h"
+#include "glm/vec2.hpp"
+
+namespace Engine {
```
Snow106: { should be in new line

Engine/src/Engine/InputSystem.h
```diff
@@ -0,0 +1,118 @@
+#pragma once
+#include <map>
+#include <mutex>
+#include "GLFW/glfw3.h"
```
Snow106: can be forward declared

thirdparty/gtest/googletest/generated/GTestConfig.cmake
```diff
@@ -1,33 +1,33 @@
-
```
Snow106, line 1: where the gtests intentional changes?
We might need to add some stuff to .gitignore

Snow106: The changes seem to leave the files identical so I wonder if it there are some vs shenanigans going on

Snow106 (changes requested):
Nice work! 🔥
Just minor changes

Engine/src/Engine/InputSystem.cpp
roughleythomas: Constructor was made private to conform to the singleton pattern.

Engine/src/Engine/InputSystem.h
roughleythomas: Was just abusing the auto comments from visual studio

Engine/src/Engine/InputSystem.h
roughleythomas: ???

thirdparty/gtest/googletest/generated/GTestConfig.cmake
roughleythomas: Never touched the cmake so probably

Snow106 (dismissed):

Snow106 (dismissed):

Engine/src/Engine/InputSystem.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+using namespace Engine;
+
+InputSystem* InputSystem::m_instance = nullptr;
+std::mutex InputSystem::m_instanceMutex;
+
+
+InputSystem* InputSystem::getInstance()
+{
+	std::unique_lock<std::mutex> lock;
```
KevinAlexV: Is there a reason unique_lock is used instead of lock_guard?

Engine/src/Engine/InputSystem.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+	if (!m_instance)
```
KevinAlexV, line 14: Would this check for nullptr, or is it just checking for NULL? I recall nullptr being a guideline thing, but if null is better in this case, that's fair.

Engine/src/Engine/InputSystem.cpp
```diff
@@ -0,0 +1,110 @@
...
+		auto cursorFunc = [](GLFWwindow* window, double x, double y)
+		{
+			static_cast<InputSystem*>(glfwGetWindowUserPointer(window))->cursorCallback(window, x, y);
+		};
+		glfwSetCursorPosCallback(window, cursorFunc);
+		m_isInit = true;
+	}
+}
+
+bool InputSystem::isButtonPressed(int button)
```
KevinAlexV, line 68: These look really clean!

Engine/src/Engine/InputSystem.h
```diff
@@ -0,0 +1,120 @@
...
+#pragma endregion
+#pragma region Callbacks
+		/// <summary>
+		/// Callback function registered with glfw
+		/// </summary>
+		/// <param name="window">Window context</param>
+		/// <param name="button">Active button</param>
+		/// <param name="action">Action of button</param>
+		/// <param name="mods">Modifier bits (shift,ctrl,etc.)</param>
+		void mouseCallback(GLFWwindow* window, int button, int action, int mods);
```
KevinAlexV, line 67: Is 'Action of Button' just representative of whether or not the button is pressed, released, etc?

Engine/src/Engine/Scene.cpp
```diff
@@ -93,8 +95,8 @@ namespace Engine
 
 
 		//Setup a callback to update the viewport size when the window is resized
-		glfwSetWindowUserPointer(m_window, reinterpret_cast<void*>(this));
```
KevinAlexV, line 96: why are these two lines commented out?

KevinAlexV (commented):
Short and sweet! Just wanted to clarify a few things, if you don't mind reading my comments.

Engine/src/Engine/InputSystem.cpp
roughleythomas: Changed it to use lock_guard

Engine/src/Engine/InputSystem.cpp
roughleythomas: Thanks

Engine/src/Engine/InputSystem.h
roughleythomas: Yes

Engine/src/Engine/Scene.cpp
roughleythomas: To register the callback function with glfw, the user pointer has to be set to an instance of InputSystem. However this would conflict with the window resize callback. Since it was apparently not being used, I commented it out to prevent conflicts. If it is required there is a work around that can be implemented quickly.

Engine/src/Engine/InputSystem.cpp
roughleythomas: This checks for both NULL and nullptr

Engine/src/Engine/Scene.cpp
Hevano: That makes sense. When we add handling for screen resizing we can apply that through the input system rather than through the scene directly.

Hevano (approved):
Everything looks good to me.

Snow106 (approved):

PR #37, opened 2022-11-06
Serialization Part 2: Electric Serialoo

Snow106:
* Added method to serialize entities individually.
* Serialization will create a json in the same format ad the sample one in previous PR
* Functionality still needs to be added to actually call serialization.
* I called serialization on our current scene, manually copied the json data, and added it as our new scene file ("testScene.bda").


Engine/src/Engine/Entity.h
```diff
@@ -62,6 +62,9 @@ namespace Engine
 		{
 			return !(*this == other);
 		}
+
+		//Returns true if entity is null
+		operator bool() const { return m_entityHandle != entt::null; }
```
KevinAlexV, line 67: We might need to clarify this in the style guidelines, but if the line of code is one line, should we keep it on the same line or still use newlines for the braces and command?

KevinAlexV (dismissed):
Just a side-note about guidelines, but this looks great!
Additionally, might want to resolve the merge conflicts - I didn't notice it until now

Engine/src/Engine/Entity.h
Snow106: we have some of these already, so I just kept it in. I think it's fine to have these in 1 line as long as it's 1 operation. We can add it to the guidelines if people are ok w it

KevinAlexV (approved):

ReleaseTheSpice (approved):

PR #41, opened 2022-11-08
Refactored vertex buffer data functions in scene; added rectangle shape

mikesinclair66:
-Refactored getSpriteVbo, getSpriteVao, getSpriteIbo into getVbo, getVao, and getIbo (so they are now general-purpose towards different vertex data)
-Unbinded vertex buffers after vertex component creation
-Added create rectangle function as alternative to create sprite


Engine/src/Engine/Scene.h
```diff
@@ -76,11 +77,14 @@ namespace Engine
...
         VerticesComponent createSprite();
+        VerticesComponent createRectangle();
+        void setupVertexAttribPtr(VerticesComponent& vc);
 		
 		//Functions to create sprite specific OpenGL buffers. Returns the ID of said buffers.
-		GLuint getSpriteVBO();
-        GLuint getSpriteVAO();
-        GLuint getSpriteIBO();
+
+		GLuint getVBO(std::string message = "sprite");
```
Snow106: we should prefer enums over strings to diferenciate between types

Game/imgui.ini
```diff
@@ -0,0 +1,30 @@
+[Window][Debug##Default]
```
Snow106, line 1: What is this file for?

Engine/src/Engine/Scene.h
KevinAlexV: If we're going to use variables to determine what kind of VBO should be generated, let's use an Enum.
It'll be easier to expand and use in the future.

Engine/src/Engine/Scene.h
```diff
@@ -76,11 +77,14 @@ namespace Engine
 
 		//Create quad for sprites
         VerticesComponent createSprite();
+        VerticesComponent createRectangle();
```
KevinAlexV, line 79: I'm still confused why we should create a Rectangle separately from a sprite. I understand wanting to create different geometric objects, but in this case the rectangle is practically identical to a sprite, just without UV coords (which aren't used anyway if a texture isn't bound). Is there a reason for this? 

Additionally, these additional object creations should wait until we create a proper renderer class to store 'renderables'

Engine/src/Engine/Scene.cpp
```diff
@@ -335,66 +380,71 @@ namespace Engine
 #pragma region Renderable Entities
 	
 	//Return the VBO for sprites. If it doesn't exist, create it.
-	GLuint Scene::getSpriteVBO() 
+	GLuint Scene::getVBO(std::string message)
```
KevinAlexV: See comments in header.

Engine/src/Engine/Scene.cpp
```diff
@@ -335,66 +380,71 @@ namespace Engine
...
 	{
-		if (!m_createdIBO);
-		{
-			m_createdIBO = true;
-			
+		GLuint ibo;
+		glGenBuffers(1, &ibo);
+		glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, ibo);
+
+		if (message == "rectangle") {
```
KevinAlexV: If we're going to have a 'sprite' and a 'rectangle' we should probably declare them separately. This comes back to the question of "why do we want two objects". 

Ultimately, if we are going to use the same indices/vertices for an object, we should store them in a variable (map or otherwise) to access the same data and not re-declare it each time, IMO.

KevinAlexV (changes requested):
This seems like a good change, so thank you for taking point on this! We'll have to refactor these into a renderer class, but having the setup should be fine. 

Few comments in regards to the current changes, can you take a look?

Engine/src/Engine/Scene.cpp
```diff
@@ -210,9 +210,16 @@ namespace Engine
...
 			}
 
-			if(m_registry.all_of<AnimationComponent>(entity))
+			if (m_registry.all_of<AnimationComponent>(entity))
 			{
-				const auto anim = m_registry.get<const AnimationComponent>(entity);
+				glBindBuffer(GL_ARRAY_BUFFER, vertices.vboID);
+				auto& anim = m_registry.get<AnimationComponent>(entity);
+				anim.deltaTime += dt;
+				if (anim.deltaTime > 0.3) {
```
Snow106: { in new line

Engine/src/Engine/Scene.cpp
```diff
@@ -232,14 +239,30 @@ namespace Engine
 				glBufferData(GL_ARRAY_BUFFER, sizeof(vertices), vertices, GL_DYNAMIC_DRAW);
 
 				glBindTexture(GL_TEXTURE_2D, anim.texID);
-				
+			}
+			else {
```
Snow106: { in new line

Engine/src/Engine/Scene.cpp
Snow106: Also 0.3 is a magic variable.

Engine/src/Engine/Scene.cpp
```diff
@@ -210,9 +210,16 @@ namespace Engine
...
+			if (m_registry.all_of<AnimationComponent>(entity))
 			{
-				const auto anim = m_registry.get<const AnimationComponent>(entity);
+				glBindBuffer(GL_ARRAY_BUFFER, vertices.vboID);
+				auto& anim = m_registry.get<AnimationComponent>(entity);
+				anim.deltaTime += dt;
+				if (anim.deltaTime > 0.3) {
+					anim.deltaTime = 0;
+					anim.currentIndex++;
+					if (anim.currentIndex > 8) anim.currentIndex = 0;
```
Snow106: 8 is a magic variable

Engine/src/Engine/Scene.h
Snow106: Enum should be used here ideally, so it's clear what the possible options are.

Engine/src/Engine/Scene.cpp
```diff
@@ -412,27 +462,46 @@ namespace Engine
...
+		vc.vaoID = getVAO();
+		vc.vboID = getVBO("sprite");
+		vc.iboID = getIBO("rectangle");
+
+		setupVertexAttribPtr(vc);
+
+		return vc;
+	}
+
+	VerticesComponent Scene::createRectangle() {
```
Snow106: { in new line

Engine/src/Engine/Scene.cpp
```diff
@@ -335,66 +380,71 @@ namespace Engine
...
 			};
-
-			glGenBuffers(1, &m_spriteIBO);
-			glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, m_spriteIBO);
-
 			glBufferData(GL_ELEMENT_ARRAY_BUFFER, sizeof(indices), indices, GL_STATIC_DRAW);
 		}
-
-		return m_spriteIBO;
+		else {
```
Snow106: { in new line

Engine/src/Engine/Scene.cpp
```diff
@@ -412,27 +462,46 @@ namespace Engine
 	VerticesComponent Scene::createSprite()
 	{
 		VerticesComponent vc;
-		
 		vc.vertexAttributes.push_back(VertexAttribute(0, 3, GL_FLOAT, GL_FALSE, 0));
 		vc.vertexAttributes.push_back(VertexAttribute(1, 2, GL_FLOAT, GL_FALSE, sizeof(float) * 3));
-		
-		//TODO: Update vertexAttributes for UV
 		vc.stride = sizeof(float) * 5;
 		vc.numIndices = 6;
```
Snow106: 5 and 6 are both magic variables

Engine/src/Engine/Scene.cpp
```diff
@@ -412,27 +462,46 @@ namespace Engine
...
+		setupVertexAttribPtr(vc);
+
+		return vc;
+	}
+
+	VerticesComponent Scene::createRectangle() {
+		VerticesComponent vc;
+		vc.vertexAttributes.push_back(VertexAttribute(0, 3, GL_FLOAT, GL_FALSE, 0));
+		vc.stride = sizeof(float) * 5;
+		vc.numIndices = 6;
```
Snow106: magic variables

Snow106 (changes requested):
there seems to be a lot of code repetition across the new classes.
I would suggest taking a look t common elements and code lines among them, and figuring out a way of putting them in a method where the parameters (or helper functions) define the different functionality.

Engine/src/Engine/Scene.cpp
mikesinclair66: Great observation. I came to this conclusion because when I initially had static variables set that can be reused (rectangle or sprite), OpenGL seems to in some way operate directly with the set of vertices and in result, only the last object gets drawn. So the only conclusion I could arrive to is redeclare the vertices per buffer object.

Game/imgui.ini
mikesinclair66: Was this a new file added? I'm oblivious to why or when this was added. Maybe it was kevin.

Snow106: this looks like a GUI thing. I think you can remove it safely from this branch, might need to reach out to the GUI team

Engine/src/Engine/Scene.h
mikesinclair66: Check our convo on discord general for this lol. Basically GL stride offset of its vbos gets set based on 5 values with sprites but only three for rectangles. In other words, the gl binded data is slower, especially considering all the tiles don't need to be bounded to a texture when they're flat colored or with a gradient. May as well make it efficient since the texture will be bounded for each instance otherwise

Engine/src/Engine/Scene.cpp
KevinAlexV: I think that was just based on how we were previously storing the vertices. If we store them in separate VBOs etc, it should fix but I'm not 100% sure on that. Worth exploring though, for sure.

Engine/src/Engine/Scene.h
KevinAlexV: Based on our convo in person, this would be based on shaders, right? Could we alter the shader file we use for this? Worth discussing in the rendering refactor

Engine/src/Engine/Scene.cpp
```diff
@@ -210,9 +209,20 @@ namespace Engine
...
 
-			if(m_registry.all_of<AnimationComponent>(entity))
+			if (m_registry.all_of<AnimationComponent>(entity))
 			{
-				const auto anim = m_registry.get<const AnimationComponent>(entity);
+				glBindBuffer(GL_ARRAY_BUFFER, vertices.vboID);
+				auto& anim = m_registry.get<AnimationComponent>(entity);
+				anim.deltaTime += dt;
+				
+				double updateTime = 0.3;
```
Snow106: I would expect 0.3 to be a frameRate variable declared in the spriteSheet component

Snow106: similar to spritesOnSheet

Engine/src/Engine/Scene.cpp
```diff
@@ -335,66 +384,72 @@ namespace Engine
...
 	{
-		if (!m_createdIBO);
-		{
-			m_createdIBO = true;
-			
+		GLuint ibo;
+		glGenBuffers(1, &ibo);
+		glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, ibo);
+
+		if (type == RenderableType::RECTANGLE) {
```
Snow106: our current rule is all caps for const global variables.
We don't have any style guidelines for enums, but in serializer, I kind of set one by grabbing two letters from enum and then the name of the enum since I've seen that convention elsewhere. In this case, it would be RT_Rectangle.
If that sounds like a good standard we can add it.

Snow106 (changes requested):

KevinAlexV (approved):
Thank you for bringing up a lot of flaws with this design! We'll take it moving forward with the renderer class

Snow106 (approved):

PR #44, opened 2022-11-09
Serialization Part 3: Revenge of the Scripth

Snow106:
Added serialization/deserialization for ScriptComponents.
* Had to change to structure of ScriptableBehavior and implemented a pure virtual method that should always return the name of the script as a string
* Added a python script that uses a template to generate a cpp class. That cpp class is a simple string comparisson that links a given scriptName to the correct script Type.
* Added as part of the python script a copy of the scripts from the game dir to the engine dir. This was due to dlls and found no alternative way to fix it.
* Updated gitignore for relavant generated and copied files.
* Updated scene.cpp and removed the 'CreateEntities' method. Moved relevant functionality to where best suited.
* Added a cmake pre_build command to run the python script before the project is built.


Engine/src/Engine/ResourceManagement/Serializer.cpp
```diff
@@ -191,7 +194,6 @@ namespace Engine
 			j["texWidthFraction"] = c.texWidthFraction;
 			j["texHeightFraction"] = c.texHeightFraction;
 			j["texName"] = c.texName;
-			j["numSprites"] = c.spritesOnSheet;
```
KevinAlexV: Why was this removed? Just because it's regenerated? Because if so, there are probably a few other variables we can remove.

KevinAlexV (commented):
Looks good! Nice to have scripts fully implemented (right?), so I'm good to approve this. 

When you get a chance though, check up on the failed build. Not too sure why it's failing.

Engine/src/Engine/ResourceManagement/Serializer.cpp
Snow106: I didn't mean to remove this. added back in

KevinAlexV (approved):
Looking good!

ReleaseTheSpice (approved):
This is some clever stuff, extremely impressive!

PR #46, opened 2022-11-14
Feature/imgui render

Aiden-Lonquist:
GUI functionality with game window rendering on same window as UI panels.


Engine/src/Engine/EngineUI/FrameBuffer.cpp
```diff
@@ -0,0 +1,52 @@
+#include "FrameBuffer.h"
```
Snow106: I would suggest leaving this class out of the PR since it's out of scope of this task.

Engine/src/Engine/EngineUI/UserInterface.h
```diff
@@ -45,14 +49,19 @@ namespace Engine
 		//partition defines the title section of the UI element
 		void partition(const std::string& fontTag, const std::string& title = "", const ImVec4& color = white);
 
-		virtual void show() = 0;
+		virtual void update() {}
+		virtual void show(unsigned int fbo = 0) = 0;
```
Snow106: do all children of UserInterface expect and fbo to passed in? What if the panel has no rendering involved?

Engine/src/Engine/HierarchyPanel.cpp
```diff
@@ -0,0 +1,71 @@
...
+		{
+			ImGui::SetWindowPos("HierarchyPanel", ImVec2(m_position.x, m_position.y));
+			isPositionSet = true;
+		}
+
+		ImGui::SetWindowSize("HierarchyPanel", ImVec2(m_dimension.x, m_dimension.y));
+
+		//Need some .otf/.ttf font files
+		//defines the title section above the UI element
+		partition("MyriadPro_Bold_16", "Entities", grey);
```
Snow106: font should be a private member here

Engine/src/Engine/HierarchyPanel.h
```diff
@@ -0,0 +1,30 @@
...
+
+	public:
+
+		virtual ~HierarchyPanel() {}
+
+		bool isAddButtonClicked() const;
+
+		const std::string& getSelectedEntity() const;
+
+		void addEntity(const std::string entity);
```
Snow106: prefer const& for string

Engine/src/Engine/InspectorPanel.cpp
```diff
@@ -0,0 +1,132 @@
...
+				ImGui::Text(name.c_str());
+
+				setSpacing(spacing);
+			}
+
+			if (m_registry->any_of<AnimationComponent>(m_entityHandle))
+			{
+				auto* animation = &(m_registry->get<AnimationComponent>(m_entityHandle));
+
+				partition("Freedom_18", "Animation", darkGreen);
```
Snow106: font should be a private member

Engine/src/Engine/SceneBuilder/Entity.cpp
```diff
@@ -2,7 +2,16 @@
 
 namespace Engine
 {
+    int Entity::s_totalEntities = 0;
+
+    int Entity::getTotalEntities()
+    {
+        return s_totalEntities;
```
Snow106: if the intention is to get the number of entities, we can just call m_registry.size();

Engine/src/Engine/SceneBuilder/Scene.h
```diff
@@ -99,14 +110,37 @@ namespace Engine
...
+        //inspector panel is the UI element that lists all the components for each entity
+        InspectorPanel m_inspectorPanel;
+
+        //game panel is the UI element that shows the render of the game
+        GamePanel m_gamePanel;
+
+        //dialog box to allow user to input the tag name for an entity
+        TagDialog m_tagDialog;
+
+        enum ComponentPanels
```
Snow106: since this is an UI specific enum, it would be better placed outside of scene.

Engine/src/Engine/SceneBuilder/Scene.h
```diff
@@ -99,14 +110,37 @@ namespace Engine
         int m_windowWidth{1920};
         int m_windowHeight{1080};
 
+        //Added this to keep track of the entities created
+        std::unordered_map<std::string, entt::entity> m_entityHandles;
```
Snow106: the registry already keeps track of entities created, and entities have an associated TagComponent

Engine/src/Engine/SceneBuilder/Scene.h
```diff
@@ -99,14 +110,37 @@ namespace Engine
...
+
+
+        //flag, if true: ui is rendered, if false: no ui is rendered
+        bool showUI = true;
```
Snow106, line 115: this should be a build flag, not a local one

Engine/src/Engine/SceneBuilder/Scene.h
```diff
@@ -24,6 +28,12 @@ namespace Engine
     public:
 
         #pragma region Entity Management
+
+        unsigned int m_fbo;
+        unsigned int textureColorbuffer;
```
Snow106: I know some of these are my commits, but those were under the impression that this branch would be discarded.
For Rendering specific things, please keep them out of the review since they're out of the scope of the PR

Engine/src/Engine/EngineUI/GamePanel.cpp
```diff
@@ -0,0 +1,49 @@
...
+		}
+
+		//defines the position and size of the entities UI element
+		ImGui::SetWindowPos("Game", ImVec2(m_position.x, m_position.y));
+		ImGui::SetWindowSize("GamePanel", ImVec2(m_dimension.x, m_dimension.y));
+
+		ImGui::End();
+
+
+		//ImGui::Begin("GamePanel", nullptr,
```
Snow106: Remove commented code

Engine/src/Engine/ResourceManagement/Serializer.cpp
```diff
@@ -127,6 +127,7 @@ namespace Engine
 
 		std::cout << sceneJson << std::endl;
 		//TODO: Bind serialization to GUI event once we have one.
+		ResourceManager::getInstance()->saveJsonFile(sceneJson, sceneFile);
```
Snow106, line 130: Not a request to change but FYI @JackyLau01 mentioned interest in changing the way these two functions behave (by returning json's and receiving it as input, instead of making use of the Resource Manager. May want to check with them to avoid future headaches.

Engine/src/Engine/EngineUI/ComponentsPanel.cpp
```diff
@@ -4,32 +4,66 @@ namespace Engine
...
+	{
+		return m_selectedComponent;
+	}
+
+	void ComponentsPanel::addComponent(const std::string& component)
+	{
+		m_components.push_back(component);
+	}
+
+	void ComponentsPanel::show(unsigned int fbo)
```
KevinAlexV: Why are we passing an FBO to this function? I don't see it being used anywhere, so I'm kinda curious what's going on.

Snow106 (changes requested):

Engine/src/Engine/HierarchyPanel.cpp
```diff
@@ -0,0 +1,71 @@
+#include "HierarchyPanel.h"
```
Snow106: Hierarchy and Inspector Panels should be in EngineUI folder

Engine/src/Engine/TagDialog.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
```
Snow106: TagDialog should be in EngineUI

Snow106 (changes requested):

Engine/src/Engine/EngineUI/EntitiesPanel.cpp
```diff
@@ -2,7 +2,7 @@
 
 namespace Engine
 {
-	void EntitiesPanel::show()
+	void EntitiesPanel::show(unsigned int fbo)
```
KevinAlexV: See first comment about FBOs

Engine/src/Engine/EngineUI/FrameBuffer.cpp
```diff
@@ -0,0 +1,52 @@
...
+		glBindRenderbuffer(GL_RENDERBUFFER, renderBufferID);
+		glRenderbufferStorage(GL_RENDERBUFFER, GL_DEPTH24_STENCIL8, width, height);
+		glBindRenderbuffer(GL_RENDERBUFFER, 0);
+		glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_RENDERBUFFER, renderBufferID);
+
+		// check for completeness
+		unsigned int completeStatus = glCheckFramebufferStatus(GL_FRAMEBUFFER);
+		if (completeStatus != GL_FRAMEBUFFER_COMPLETE) 
+		{
+			//error
```
KevinAlexV: Should probably write an error to the engine log

Engine/src/Engine/EngineUI/FrameBuffer.cpp
```diff
@@ -0,0 +1,52 @@
...
+			//error 
+		}
+
+		glBindFramebuffer(GL_FRAMEBUFFER, 0);
+	}
+
+	FrameBuffer::~FrameBuffer()
+	{
+		glDeleteFramebuffers(1, &fbo);
+		fbo = 0;
```
KevinAlexV: Do these need to be set to 0 when destructing a frame buffer? I would think the entire class would be deleted, so setting this to 0 seems redundant.

Engine/src/Engine/EngineUI/FrameBuffer.h
```diff
@@ -0,0 +1,25 @@
...
+		inline unsigned int GetRenderBufferID() const { return renderBufferID; }
+
+
+	private:
+		unsigned int fbo;
+		unsigned int textureID;
+		unsigned int renderBufferID;
+
+		float windowWidth, windowHeight;
+		float clearColorRed, clearColorGreen, clearColorBlue, clearColorAlpha;
```
KevinAlexV: Should add 'm_' for style guidelines

Engine/src/Engine/EngineUI/GamePanel.cpp
```diff
@@ -0,0 +1,49 @@
...
+namespace Engine
+{
+	void GamePanel::show(unsigned int fbo)
+	{
+
+		ImGui::Begin("GamePanel");
+		{
+			ImGui::BeginChild("GamePanel");
+			ImVec2 wsize = ImGui::GetWindowSize();
+			ImGui::Image((ImTextureID)fbo, ImVec2(1920, 1080), ImVec2(0, 1), ImVec2(1, 0));
```
KevinAlexV, line 12: So the FBO is used for the game panel to be rendered on screen? Does that mean only the game panel needs to be put into the FBO or is there more to this? I need to learn before I push out a renderer refactor 😅

Engine/src/Engine/EngineUI/MainMenu.cpp
```diff
@@ -2,12 +2,14 @@
 
 namespace Engine
 {
-	void MainMenu::show()
+	void MainMenu::show(unsigned int fbo)
```
KevinAlexV, line 5: See other comment about FBO

Engine/src/Engine/HierarchyPanel.cpp
```diff
@@ -0,0 +1,71 @@
...
+			{
+				if (ImGui::TreeNode(entity.c_str()))
+				{
+					m_selectedEntity = entity;
+					ImGui::TreePop();
+				}
+			}
+		}
+
+		ImGui::End();
```
KevinAlexV: I always like seeing how clean this is, well done!

Engine/src/Engine/InspectorPanel.h
```diff
@@ -0,0 +1,32 @@
...
+		void setRegistry(entt::registry* registry);
+
+		void setSelectedEntity(entt::entity entityHandle);
+
+		virtual void show(unsigned int fbo = 0) override;
+
+	private:
+
+		entt::entity m_entityHandle {0};
+		entt::registry* m_registry {nullptr};
```
KevinAlexV: Correct me if I'm wrong, but shouldn't this pointer be deleted upon the destruction of the InspectorPanel class? I don't see that anywhere.

Engine/src/Engine/InspectorPanel.cpp
```diff
@@ -0,0 +1,132 @@
...
+				std::string id = "ID: " + std::to_string(texture->texID);
+				std::string name = "Name: " + texture->texName;
+
+				ImGui::Text(id.c_str());
+				ImGui::Text(name.c_str());
+
+				setSpacing(spacing);
+			}
+
+			if (m_registry->any_of<AnimationComponent>(m_entityHandle))
```
KevinAlexV: Seeing my components being used brings a tear to my eye. Really clean and easy to interpret - well done!

Engine/src/Engine/InspectorPanel.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+				std::string index = "Index: " + std::to_string(animation->currentIndex);
+				std::string numPerRow = "Number per row: " + std::to_string(animation->numPerRow);
+
+				ImGui::Text(index.c_str());
+				ImGui::Text(numPerRow.c_str());
+
+				ImGui::SliderFloat("Frame rate", &animation->frameRate, 0.0f, 1.0f, "%.2f");
+				ImGui::SliderFloat("Texture width", &animation->texWidthFraction, 0.0f, 1.0f, "%.2f");
+				ImGui::SliderFloat("Texture height", &animation->texHeightFraction, 0.0f, 1.0f, "%.2f");
```
KevinAlexV: When the render refactor is complete, I'll likely come back to this to adjust the names, as they may be inaccurate. For now though, this is good

(Though I do think a better name for the tex width/height would be sprite frame width/height. It's the % of the spriteSheet/texture that each sprite falls under)

Engine/src/Engine/SceneBuilder/Scene.cpp
```diff
@@ -28,11 +29,62 @@ namespace Engine
...
+
+		glGenVertexArrays(1, &quadVAO);
+		glGenBuffers(1, &quadVBO);
+		glBindVertexArray(quadVAO);
+		glBindBuffer(GL_ARRAY_BUFFER, quadVBO);
+		glBufferData(GL_ARRAY_BUFFER, sizeof(quadVertices), &quadVertices, GL_STATIC_DRAW);
+		glEnableVertexAttribArray(0);
+		glVertexAttribPointer(0, 2, GL_FLOAT, GL_FALSE, 4 * sizeof(float), (void*)0);
+		glEnableVertexAttribArray(1);
+		glVertexAttribPointer(1, 2, GL_FLOAT, GL_FALSE, 4 * sizeof(float), (void*)(2 * sizeof(float)));
```
KevinAlexV: This is a HUGE chunk of code that I feel like can be functioned off. However, with the render refactor, this can probably be left in for now - I just likely will reach out for assistance in refactoring this.

Engine/src/Engine/SceneBuilder/Scene.h
KevinAlexV: 'm_' for style guidelines. For textureColorbuffer, and quadVAO/quadVBO.

Game/imgui.ini
```diff
@@ -5,7 +5,7 @@ Collapsed=0
 
 [Window][ExplorerPanel]
```
KevinAlexV: Out of curiosity, should this ini be added to the gitignore? I noticed it gets generated each time the engine is run, and/or modified. But this also seems client-specific, and not something everyone would share the properties of, if that makes sense. 

Would love a better understanding though!

KevinAlexV (commented):
Looks good guys, great job! I believe after this, we need to add functionality, right? 

Few comments down below, if you can take a look. 

Finally: Can you also make sure the classes/headers you created are within the proper folders? The stuff like HierarchyPanel and InspectorPanel should be under the EngineUI folder. Tag dialog, if only used for the Engine UI, should also be moved there.

Engine/src/Engine/EngineUI/UserInterface.h
Aiden-Lonquist: You're right, the others did not need to be passed fbo. It was there for testing but I forgot to remove it

Engine/src/Engine/SceneBuilder/Scene.h
Aiden-Lonquist: I may be misunderstanding, but the m_fbo is being used in this pr. The pr includes functionality for the GUI and the render panel that uses the fbo.

Engine/src/Engine/EngineUI/ComponentsPanel.cpp
Aiden-Lonquist: Accidently left there from testing. Removed

Engine/src/Engine/EngineUI/FrameBuffer.cpp
Aiden-Lonquist: This class has been removed now.

Engine/src/Engine/EngineUI/GamePanel.cpp
Aiden-Lonquist: That's right, only the game panel uses it. Sorry for the confusion.

Engine/src/Engine/SceneBuilder/Scene.cpp
Aiden-Lonquist: I agree this could be refactored. I've left it for now but can absolutely assist when we need to refactor it.

Engine/src/Engine/EngineUI/FrameBuffer.h
Aiden-Lonquist: Ended up removing this class.

Engine/src/Engine/ResourceManagement/Serializer.cpp
Aiden-Lonquist: Good to know, thanks.

Engine/src/Engine/SceneBuilder/Scene.h
Aiden-Lonquist: Right, how do I set it as a build flag?

Game/imgui.ini
Aiden-Lonquist: As far as I know, it would be fine adding it to gitignore. It's user specific and has to do with the position of the GUI panels as well as other non critical attributes.

Snow106: if that's the case lets remove the file in this pr and update the .gitignore then

Engine/src/Engine/SceneBuilder/Scene.h
Snow106: we can leave it as is for now, but add a todo on top so we can change it later. Please also create an associated trello ticket to change this to build flags

Aiden-Lonquist: sounds good

Game/imgui.ini
Aiden-Lonquist: removed the file

Engine/src/Engine/SceneBuilder/Scene.h
Aiden-Lonquist: fbo stuff is now removed from this pr. The game window is still built to take in an fbo but uses a default value of 0 for the time being.

Engine/src/Engine/HierarchyPanel.cpp
Blackstaxs: fixed

Engine/src/Engine/InspectorPanel.cpp
Blackstaxs: fixed

Engine/src/Engine/InspectorPanel.h
Blackstaxs: No, this is only a reference to the registry, which is kept in the scene. The Inspector has no power over destroying this, it only references it.

Engine/src/Engine/SceneBuilder/Scene.h
Blackstaxs: I moved it to ComponentsPanel.cpp

Engine/src/Engine/SceneBuilder/Entity.cpp
Blackstaxs: fixed

Engine/src/Engine/SceneBuilder/Scene.h
Blackstaxs: fixed

Snow106 (approved):
good work!

Engine/src/Engine/SceneBuilder/Scene.h
Aiden-Lonquist: line has been removed from this PR

KevinAlexV (approved):

PR #47, opened 2022-11-14
Feature/advanced shader

mikesinclair66:
shaders are loaded via vertices component. Also, shaders aren't loaded per frame; when a component of separate vertex data is being loaded, its corresponding shader is assigned.


Engine/src/Engine/SceneBuilder/Scene.h
```diff
@@ -4,6 +4,7 @@
 #include "Components.h"
 #include <Engine/Utilities/DeltaTime.h>
 #include "Engine/Core.h"
+#include <Engine/ResourceManagement/ShaderGenerator.h>
```
Snow106: This can be forward declared

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
```diff
@@ -1,6 +1,13 @@
 #pragma once
 #ifndef SHADERGENERATOR_H
 #define SHADERGENERATOR_H
+
+#include "Engine/ResourceManagement/ResourceManager.h"
+#include <string>
+#include <map>
+
+using namespace std;
```
Snow106: guidelines: no using namespace in header files

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
```diff
@@ -1,6 +1,13 @@
 #pragma once
 #ifndef SHADERGENERATOR_H
 #define SHADERGENERATOR_H
+
+#include "Engine/ResourceManagement/ResourceManager.h"
```
Snow106: I don't see this class being used in this header

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
```diff
@@ -11,14 +18,28 @@ namespace Engine
...
 		int m_programId;
 		int genShader(const char* source, GLenum shaderType);
 		bool requestStatus(GLuint shaderObj, GLenum status);
 	};
+
+	class ShaderNorms {
+	public:
+		ShaderNorms();
+		~ShaderNorms();
+		void addShader(string);
```
Snow106: this isn't in the style guidelines, but I believe we've been consisting with parameter names in function signatures as well as definitions. Lets keep that consistent

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
```diff
@@ -11,14 +18,28 @@ namespace Engine
...
-		~ShaderGenerator();
 		int getProgramId() const;
 
 	private:
 		int m_programId;
 		int genShader(const char* source, GLenum shaderType);
 		bool requestStatus(GLuint shaderObj, GLenum status);
 	};
+
+	class ShaderNorms {
```
Snow106: should we have this class in a separate file?

Engine/src/Engine/ResourceManagement/ShaderGenerator.cpp
```diff
@@ -60,4 +55,50 @@ namespace Engine
...
+
+	ShaderNorms::~ShaderNorms()
+	{
+		for (auto [key, value] : m_shaders)
+		{
+			glDeleteProgram(value);
+		}
+	}
+
+	void ShaderNorms::addShader(string shaderName)
```
Snow106: prefer const string&

Engine/src/Engine/ResourceManagement/ShaderGenerator.cpp
```diff
@@ -60,4 +55,50 @@ namespace Engine
...
+		if (stride != curStride)
+		{
+			curStride = stride;
+			changed = true;
+		}
+
+		return changed;
+	}
+
+	GLuint ShaderNorms::getShader(string shaderName)
```
Snow106: prefer const string &

Engine/src/Engine/ResourceManagement/ShaderGenerator.cpp
```diff
@@ -60,4 +55,50 @@ namespace Engine
...
+	{
+		return m_shaders.at(shaderName);
+	}
+
+	GLuint ShaderNorms::getDefaultShader()
```
Snow106: not a request change, but it feels like this could just be an overload of getShader when no arguments are passed.

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
```diff
@@ -11,14 +18,28 @@ namespace Engine
...
+		~ShaderNorms();
+		void addShader(string);
+		bool queueStride(int);
+		GLuint getShader(string);
+		GLuint getDefaultShader();
+
+	private:
+		map<string, GLuint> m_shaders;
+		string m_defaultShader;
+		int curStride{ -1 };
```
Snow106: guideline: should have m_ prefix

Snow106: also, I would suggest avoiding abbreviations here, it's not as immediately obvious as just reading currentStride

Engine/src/Engine/ResourceManagement/ShaderGenerator.cpp
```diff
@@ -60,4 +55,50 @@ namespace Engine
...
+	{
+		m_shaders.insert(pair(shaderName, ShaderGenerator(
+			ResourceManager::getInstance()->getShaderData(shaderName + ".vs").c_str(),
+			ResourceManager::getInstance()->getShaderData(shaderName + ".fs").c_str()
+		).getProgramId()));
+		GE_CORE_INFO(shaderName +" ADDED.");
+	}
+
+	//true if a new stride amount is introduced so that GL doesn't reload the same shader
+	bool ShaderNorms::queueStride(int stride)
```
Snow106: since this returns a bool I think the naming of this method should be a bit more informative. Something like isNewStride, or something to that effect

Engine/src/Engine/ResourceManagement/ShaderGenerator.cpp
```diff
@@ -60,4 +55,50 @@ namespace Engine
 		}
 		return true;
 	}
+
+	ShaderNorms::ShaderNorms() : m_defaultShader("TextureFill")
```
Snow106: would it be possible to have an enum for these shaders instead of a string map?

Engine/src/Engine/SceneBuilder/Scene.cpp
```diff
@@ -269,6 +266,19 @@ namespace Engine
 				glBufferData(GL_ARRAY_BUFFER, sizeof(vertices), vertices, GL_DYNAMIC_DRAW);
 			}
 
+			//load shader according to vertices component (given shader is not already loaded)
+			if (shaderNorms != nullptr && shaderNorms->queueStride(vertices.stride))
```
Snow106: this feels like it can be a function encapsulated in the ShaderNorms. Would help since we're trying to trim and decouple Scene

Snow106 (changes requested):
Good work, just some changes

Engine/src/Engine/SceneBuilder/Scene.cpp
mikesinclair66: You're correct that it can be encapsulated. I don't think it's very necessary though, the statement is very dependent on members of scene such as the vertices component and program Id as well as a growing list of indices length registers. The operation seems more dependent on scene than ShaderNorms.

Engine/src/Engine/ResourceManagement/ShaderGenerator.cpp
mikesinclair66: Yes!

Engine/src/Engine/ResourceManagement/ShaderGenerator.cpp
mikesinclair66: Done.

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
mikesinclair66: Done

Engine/src/Engine/ResourceManagement/ShaderGenerator.cpp
mikesinclair66: Done

Engine/src/Engine/ResourceManagement/ShaderGenerator.cpp
mikesinclair66: Changed to enum

Engine/src/Engine/ResourceManagement/ShaderGenerator.cpp
mikesinclair66: Changed to enum

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
mikesinclair66: Done

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
mikesinclair66: Done

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
mikesinclair66: It's pretty hidden, but it needs this to access GL's GLuint which is first stored here.

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
mikesinclair66: Done

Engine/src/Engine/SceneBuilder/Scene.h
mikesinclair66: What do you mean by this? Scene.h needs to access it because it has a singleton shaderNorms object. In ShaderNorms.h now.

Engine/src/Engine/ResourceManagement/ShaderNorms.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+#ifndef SHADERNORMS_H
+#define SHADERNORMS_H
+
+#include "Engine/ResourceManagement/ResourceManager.h"
+#include <string>
+#include <map>
+
+namespace Engine {
```
Snow106: brackets in file

Engine/src/Engine/ResourceManagement/ShaderNorms.h
```diff
@@ -0,0 +1,39 @@
...
+		//shader tokens to access map of shaders
+		static enum ShaderName {
+			SN_TEXTURE_FILL,
+			SN_COLOR_FILL
+		};
+		static const ShaderName DEFAULT_SHADER_NAME;
+
+		//requires the stride of the default shader's vertices component
+		ShaderNorms(int defaultStride);
+		~ShaderNorms();
```
Snow106, line 24: rule of 0

Engine/src/Engine/ResourceManagement/ShaderNorms.cpp
```diff
@@ -0,0 +1,64 @@
+#include <Engine/ResourceManagement/ShaderNorms.h>
+#include <Engine/ResourceManagement/ShaderGenerator.h>
+#include "glad/glad.h"
+
+using namespace std;
+
+namespace Engine {
```
Snow106: fix brackets in file

Engine/src/Engine/SceneBuilder/Scene.h
mikesinclair66: Done

Snow106 (changes requested):

Engine/src/Engine/ResourceManagement/ShaderNorms.cpp
mikesinclair66: Done

Engine/src/Engine/ResourceManagement/ShaderNorms.h
mikesinclair66: Done

Engine/src/Engine/ResourceManagement/ShaderNorms.h
mikesinclair66: Done

Engine/src/Engine/ResourceManagement/ShaderNorms.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+#ifndef SHADERNORMS_H
+#define SHADERNORMS_H
+
+#include "Engine/ResourceManagement/ResourceManager.h"
```
Snow106: This is unused in the header file as far as I can see

Engine/src/Engine/SceneBuilder/Scene.h
Snow106: the include is still there after new commit

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
Snow106: We can remove this include then and forward declare the GLuint since it's what is being used in this header file

KevinAlexV (dismissed):
Looks good to me!

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
```diff
@@ -1,17 +1,18 @@
 #pragma once
 #ifndef SHADERGENERATOR_H
 #define SHADERGENERATOR_H
+
+#include <Engine/ResourceManagement/ResourceManager.h>
```
Snow106: this include is still unused

Snow106 (changes requested):

Engine/src/Engine/ResourceManagement/ShaderNorms.h
mikesinclair66: Needed to bring in the GLuint

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
mikesinclair66: Needed to bring in the GLenum. Otherwise system crashes.

Engine/src/Engine/ResourceManagement/ShaderNorms.h
Snow106: you can bring in GLuint then. Don't need to bring an entire header file just for a typedef that doesn't even belong to that header file.

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
Snow106: Same as above

Engine/src/Engine/ResourceManagement/ShaderNorms.h
Snow106: You can also just forward declare the GLuint here, like other classes in our system do

mikesinclair66: It already is forward declared if you take a look. But if the include isn't declared it crashes

Snow106: Tested on my end by adding a GLuint typedef, like recommended above, and it works fine.

Engine/src/Engine/ResourceManagement/ShaderGenerator.h
mikesinclair66: Done

Engine/src/Engine/ResourceManagement/ShaderNorms.h
mikesinclair66: Done

PR #49, opened 2022-11-16
Render Refactor

KevinAlexV:
What's been done so far for the renderer class. 

This refactor includes
-Seperating the Renderer from scene (draw calls are no longer in scene)
-Window management handling is in it's own class
-Geometry is generated in a factory (currently not complete, and temp functions are in place)


mikesinclair66 (changes requested):
Hey Kevin, so I'm still noticing the issue with the renderables appearing white. I've sent you reference, but it appears that it's due to the desctructor of my old ShaderGenerator class. Perhaps I should throw my changes into the branch and keep you updated? Otherwise, simply removing the desctructor from ShaderGenerator should do the trick. It might be better that way anyways so we keep the PR's separate.

Engine/src/Engine/Rendering/GeometryFactory.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+#pragma region Buffer Generation
+	//Return geometry based on provided geometry type.
+	VerticesComponent GeometryFactory::getVerticesComponent(Geometry geometry)
+	{
+		return defaultGeometry[geometry];
+	}
+
+	//Return the VBO for sprites. If it doesn't exist, create it.
+	GLuint GeometryFactory::getVBO(float vertices[], int verticesSize, GLsizei stride, std::vector<VertexAttribute> vertexAttributes)
```
Snow106: probably want to pass vertexattribs by ref

Engine/src/Engine/Rendering/GeometryFactory.h
```diff
@@ -0,0 +1,56 @@
...
+		#pragma endregion
+	private:
+
+		#pragma region Singleton Management
+		// The singleton instance
+		static GeometryFactory* m_instance;
+		GeometryFactory();
+		#pragma endregion
+
+		std::unordered_map<Geometry, VerticesComponent> defaultGeometry;
```
Snow106: guidelines: missing m_ prefix

Engine/src/Engine/Rendering/Renderer.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+		//Set the currentBound textures to 0 for this draw call.
+		int currentBoundTextures = 0;
+
+		//For each entitiy with a vertices component, render it.
+		for (auto [entity, vertices] : renderables.each())
+		{
+			//Set a default transform component and color if the object does not contain one.
+			TransformComponent transform{ glm::vec3 {0.f,0.f,0.f}, glm::vec3 {1.f,1.f,1.f} , glm::vec3 {0.f,0.f,0.f} };
+			glm::vec4 color{ 1.f,1.f,1.f,1.f };
```
Snow106, line 169: we probably want to use of the default constructor here instead of hardcoding these values

Engine/src/Engine/Rendering/Renderer.cpp
```diff
@@ -0,0 +1,255 @@
...
+			//Bind color, texture and shader if entity contains material.
+			if (scene.m_registry.all_of<MaterialComponent>(entity))
+			{
+				const auto material = scene.m_registry.get<const MaterialComponent>(entity);
+				
+				if (setTexture(material.texID, currentBoundTextures)){currentBoundTextures++;}
+			
+				//Change color and shaderProgram to material components color and shader.
+				color = material.color;
+				//glUseProgram(material.shaderID);
```
Snow106, line 186: should we remove this comment comment?

Engine/src/Engine/Rendering/Renderer.h
```diff
@@ -0,0 +1,65 @@
...
+		
+		//Renderer is a singleton, so this is the only instance of it.
+		static Renderer* m_pinstance;
+
+		//Define window and UI that is managed by this singleton
+		Window m_window;
+		UIRenderer m_UI;
+
+		//flag, if true: ui is rendered, if false: no ui is rendered
+		bool showUI = true;
```
Snow106: missing prefix

Engine/src/Engine/ResourceManagement/ResourceManager.h
```diff
@@ -39,27 +39,27 @@ namespace Engine
...
 
 		#pragma endregion
 
 		#pragma region Get & Set Functions
+		//Function to save provided json data to a provided file. Will save to "Assets/" directory if path is not specified.
+		void saveJsonFile(nlohmann::json data, std::string fileName, std::string path = "Assets/", std::string extension = ".json");
 
 		//Function to set icon for the application.
-		void setAppIcon(GLFWwindow&);
+		void setAppIcon(std::string appIcon, GLFWwindow* window);
```
Snow106: prefer & for string

Engine/src/Engine/EngineUI/GamePanel.h
```diff
@@ -3,13 +3,12 @@
 
 namespace Engine
 {
+	class Window;
```
ReleaseTheSpice, line 6: Could we get some comments on these classes and their methods?  Just how they are used/called and what they do is fine, but currently there are no comments in the GamePanel files

Engine/src/Engine/ResourceManagement/ScriptSerializer.cpp
```diff
@@ -0,0 +1,16 @@
+#include <string>
```
Snow106, line 1: not sure how this file got added, it's explicitly listed in the .gitignore.

Engine/src/Engine/Scripts/MoveRightScript.h
```diff
@@ -0,0 +1,13 @@
+#include "Engine/Scripts/ScriptableBehavior.h"
```
Snow106, line 1: We can remove this from this PR, it was just an example script that doesn't really serve a purpose

Engine/src/Engine/EngineUI/GamePanel.cpp
```diff
@@ -1,22 +1,27 @@
...
 {
-	void GamePanel::show(unsigned int fbo)
+	void GamePanel::setInitialPosition()
+	{
+		//defines the position and size of the entities UI element
+		ImGui::SetWindowPos("Game", ImVec2(m_position.x, m_position.y));
+		ImGui::SetWindowSize("GamePanel", ImVec2(m_dimension.x, m_dimension.y));
+	}
+
+	void GamePanel::show(const Window& window)
```
ReleaseTheSpice, line 14: Any Comments for this function?  I'm guessing it's for loading the game into center of the engine UI?

Engine/src/Engine/Utilities/DeltaTime.h
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#include "GLFW/glfw3.h"
```
Snow106, line 2: Why was this added here?

Snow106 (changes requested):

mikesinclair66 (dismissed):
Looks good

Engine/src/Engine/Utilities/DeltaTime.h
KevinAlexV: Delta time broke when it wasn't included. If it happens to work on other machines, then we can remove it - it just didn't on mine

Engine/src/Engine/Rendering/Renderer.cpp
KevinAlexV: Since shaders are being worked on, I just wanted to leave this here as a reminder. If it's an issue, I can remove it yeah

ReleaseTheSpice (approved):

Engine/src/Engine/EngineUI/GamePanel.cpp
ReleaseTheSpice: After reading through all the files, the show methods make a lot more sense., so comments here aren't necessary.  Then again, they would still make it more readable.

Engine/src/Engine/EngineUI/GamePanel.h
KevinAlexV: That's a question for the imGUI team, as they are more equipped to explain this than I am, unfortunately. We can look into it later, if needed!

Engine/src/Engine/EngineUI/GamePanel.cpp
KevinAlexV: Same as above

Snow106 (approved):

PR #50, opened 2022-11-16
Bugs/debug.exe

JianmingChen:
Fixed the issue when running game.exe cannot be loaded in debug version because the font file is not located properly.


KevinAlexV (approved):
Thanks for following up on this!

Snow106 (approved):
lgtm

PR #51, opened 2022-11-17
Added SerializableComponent to ensure runtime entities don't get serialized.

Snow106:
Use case is questionable but just looking for the worst case scenario where run time entities get serialized for no reason.


Aiden-Lonquist (approved):
Looks good.

JackyLau01:
LGTM

JackyLau01 (approved):

PR #53, opened 2022-11-19
Added shader norms

mikesinclair66:
Assigns shader based on current renderable stride. Only updates loaded shader when a new stride is assigned. Functionable for textures, flat colors, and gradients.


Engine/src/Engine/Rendering/Renderer.cpp
```diff
@@ -225,12 +228,8 @@ namespace Engine
 	//Load default shader to 'fill' object. This should be changed when multiple shaders are utilized *******************
```
ReleaseTheSpice, line 228: I take it multiple shaders are not currently utilized?

Engine/src/Engine/ResourceManagement/ShaderNorms.cpp
```diff
@@ -0,0 +1,105 @@
...
+	ShaderNorms* ShaderNorms::getInstance()
+	{
+		if (m_singleton == nullptr)
+		{
+			m_singleton = new ShaderNorms();
+		}
+		return m_singleton;
+	}
+
+	void ShaderNorms::update(int stride, int& textureCoordinates, int& colorCoordinates,
```
ReleaseTheSpice, line 41: This function is pretty important, right? It would be great to get some comments above the function itself saying what it does, and how/if you might call it as an engine developer

Engine/src/Engine/ResourceManagement/ShaderNorms.cpp
```diff
@@ -0,0 +1,105 @@
...
+		switch (shaderName)
+		{
+		case ShaderName::SN_TEXTURE_FILL:
+			return "TextureFill";
+			break;
+		case ShaderName::SN_COLOR_FILL:
+			return "ColorFill";
+			break;
+		default:
+			return "GradientFill";
```
ReleaseTheSpice, line 103: Why is gradient the default?  Do all shaders implement gradients or something like that?  I feel like texture fill is a more common case, but maybe this has something to do with how this function usually gets called?

Engine/src/Engine/Rendering/Renderer.cpp
mikesinclair66: Not utilized, no but they are available. So far the texture fill shader is used, but there's also the option to fill out a vbo with a flat color or fill a gradient.

Engine/src/Engine/ResourceManagement/ShaderNorms.cpp
mikesinclair66: It uses the default keyword because switch statements view default as "other option". Texture is indeed the default option (DEFAULT_SHADER_NAME in shadernorms), but this method just returns the string version of the corresponding enums so that the file name can be loaded.

Engine/src/Engine/ResourceManagement/ShaderNorms.cpp
mikesinclair66: Done!

ReleaseTheSpice (approved):

KevinAlexV (approved):
LGTM!

PR #61, opened 2022-11-24
Text Component and Rendering 

KevinAlexV:
Adds the Text component, which allows for text rendering on entities.


Engine/src/Engine/SceneBuilder/Components.h
```diff
@@ -15,6 +15,20 @@ namespace Engine
 {
 	class ScriptableBehavior;
 
+	//Not component, just container for vertex attribute data format
+	struct VertexAttribute
```
Snow106, line 19: can we move this just above the verticesComponent so it's a bit more clear/organized?

Engine/src/Engine/ResourceManagement/Text.h
```diff
@@ -0,0 +1,44 @@
...
+			glm::ivec2 Size;      //Size of glyph
+			glm::ivec2 Bearing;   //offset from baseline to left/top of glyph	
+			unsigned int Advance;   //Offset to advance to next glyph
+		};
+
+		//generate a texture and store its relevant data into a Character struct 
+		//that we add to the Characters map, all data require to render for later use
+		std::map<char, Character> Characters;
+	private:
+		std::string defaultFont = "arial.ttf";
```
Snow106: we might want a font enum for all our available fonts

Engine/src/Engine/ResourceManagement/ShaderGenerator.cpp
```diff
@@ -51,7 +51,7 @@ namespace Engine
 		if (!success) 
 		{
 			glGetShaderInfoLog(shaderObj, 512, nullptr, infoLog);
-			GE_CORE_ERROR("Error in shader compilation {0}", infoLog);
+			GE_CORE_ERROR("[ShaderGenerator] Error in shader compilation {0}", infoLog);
```
Snow106, line 54: I like this. We should do this.

Engine/src/Engine/Rendering/Renderer.h
```diff
@@ -57,6 +59,9 @@ namespace Engine
 		//Draw entities with a VerticesComponent to screen/FBO
 		void drawEntities(Scene& scene);
 
+		//Render text given a text component
+		void renderText(TextComponent text, TransformComponent transform, glm::vec3 color, GLuint shader, GLuint mvpID);
```
Snow106: can we pass these params, other than the GLuints, by reference instead?

Engine/src/Engine/Rendering/Renderer.cpp
```diff
@@ -253,7 +289,63 @@ namespace Engine
...
+		{
+			ch = m_text.Characters[c];
+
+			if (c != ' ')
+			{
+				charPos.scale.x = ((charPos.position.x + ch.Size.x) - (charPos.position.x + (ch.Size.x / 2)));//(target - ((baseMin * objScale) + objPos)) / baseDifference;//(1.f/ch.Size.x);
+				charPos.scale.y = ((charPos.position.y + ch.Size.y) - (charPos.position.y + (ch.Size.y / 2)));//(1.f/ch.Size.y);
+
+				charPos.position.x += ch.Bearing.x;// *scale.x;
+				charPos.position.y += 5.f * ((ch.Size.y - ch.Bearing.y) / (charPos.scale.y));// * scale.y;
```
Snow106: why is it being scaled by 5?

Engine/src/Engine/Rendering/Renderer.cpp
```diff
@@ -253,7 +289,63 @@ namespace Engine
...
+
+	//Render a string line of text based on the transform location. Based on https://learnopengl.com/In-Practice/Text-Rendering
+	void Renderer::renderText(TextComponent text, TransformComponent transform, glm::vec3 color, GLuint shader, GLuint mvpID)
+	{
+		TransformComponent charPos = transform;
+
+		//Access the current shader data/program and pass the color XYZ to the textColor variable in the shader.
+		glUniform3f(glGetUniformLocation(shader, "textColor"), color.x, color.y, color.z);
+
+		//glActiveTexture(GL_TEXTURE0);
```
Snow106: is this meant to be commented?

Engine/src/Engine/Rendering/Renderer.cpp
```diff
@@ -221,6 +226,37 @@ namespace Engine
...
+		const auto textRenderables = scene.getEntities<const TextComponent>();
+		
+		glUseProgram(m_text.m_textShaderProgram);
+		GLuint mvpID = glGetUniformLocation(m_text.m_textShaderProgram, "mvp");
+
+		for (auto [entity, text] : textRenderables.each())
+		{
+			//Set a default transform component and color if the object does not contain one.
+			TransformComponent transform;
+			glm::vec4 color{ .5f,.5f,.5f,1.f };
```
Snow106: magic variable

Engine/src/Engine/Rendering/Renderer.cpp
```diff
@@ -221,6 +226,37 @@ namespace Engine
...
+
+			//Change the transform component if the entity contains one.
+			if (scene.m_registry.all_of<TransformComponent>(entity)) { transform = scene.m_registry.get<const TransformComponent>(entity); }
```
Snow106: can we assume this is always the case since an entity with text should also have a transform?

Engine/CMakeLists.txt.bak
```diff
@@ -0,0 +1,33 @@
+file (GLOB_RECURSE sources CONFIGURE_DEPENDS "*.cpp" "*.h")
```
Snow106: not sure what generated this file but .bak files are not needed since we're using git.

Snow106 (changes requested):

Engine/src/Engine/Rendering/Renderer.cpp
KevinAlexV: Entities with the text component don't necessarily need a transform component. I wanted to remove that dependency given the nature of ECS, so I have a default transform initialized above. This is just on the chance that there is a transform component.

Engine/CMakeLists.txt.bak
KevinAlexV: Not sure either. I'll delete it

Engine/src/Engine/Rendering/Renderer.cpp
KevinAlexV: It just doesn't seem to be needed. I'll remove it for now, but it's worth keeping in mind if we ever look at batch rendering.

Engine/src/Engine/Rendering/Renderer.h
KevinAlexV: done

KevinAlexV: Done

Engine/src/Engine/ResourceManagement/Text.h
KevinAlexV: I have an idea for this, I'll see if I can quickly implement it

Engine/src/Engine/Rendering/Renderer.cpp
KevinAlexV: More to do with the scale of the text, where it needs to be scaled based on it's size. I'll try and alter it based on the character size and scale factor from transforms

Engine/src/Engine/Rendering/Renderer.cpp
KevinAlexV: added m_defaultColor to control this

Snow106 (dismissed):

PR #67, opened 2022-12-01
Fixed scene swapping

ReleaseTheSpice:


JackyLau01 (dismissed):

Engine/src/Engine/SceneBuilder/Scene.cpp
```diff
@@ -74,10 +74,25 @@ namespace Engine
 	//Per scene update loop, add scripts to entities if enabled and render the scene.
 	void Scene::onRuntimeUpdate(const DeltaTime& dt)
 	{
+		if (m_switch)
+		{
+			m_switch = false;
+			
+			if (score > 5) {
```
Snow106: braces

Engine/src/Engine/SceneBuilder/Scene.cpp
```diff
@@ -74,10 +74,25 @@ namespace Engine
...
+			}
+			
+			// Swap to scene by index score
+			swapScene(m_sceneList[score-1]);
+			
+		}
+
 		//get a view on entities with a script Component, and execute their onUpdate.
 		const auto entities = getEntities<ScriptComponent>();
+		int count = 0;
```
Snow106: we can get rid of this

Engine/src/Engine/SceneBuilder/Scene.cpp
```diff
@@ -104,11 +119,45 @@ namespace Engine
...
+			{
+				script.destroyScript();
+			}
+		}
+		
 		m_registry = entt::registry();
 		Serializer::tryDeserializeScene(*this, other);
+		
+		delete m_physics;
+		glm::vec3 dimensions;
```
Snow106, line 102: let's make this into a method since it's called in two places

Snow106 (changes requested):

roughleythomas (dismissed):

Snow106 (dismissed):

Snow106 (approved):

roughleythomas (approved):

PR #71, opened 2022-12-06
[Bugfix] Added Effects to other levels

Snow106:
Changed effect to take in a deltaTime
time in advanced shader is no longer exponential (it adds only once)
Added lerp-ish thing to effect so it slows down and matches the static tiles
This caused the effect to change slightly 🤷


KevinAlexV (dismissed):
Lookin good

KevinAlexV (dismissed):

Hevano (dismissed):
Looks good, just in time too

KevinAlexV (approved):
I like it

Aiden-Lonquist (approved):
Great work :)

PR #74, opened 2023-05-23
[Bugfix] Update Renderer.cpp

KevinAlexV:
Removed some multi-threading code, as it was causing issues for some builds.


Hevano (approved):
Thanks for fixing this

Snow106 (approved):

